PR #31, opened 2021-07-02
Sleep for 5 Seconds added while checking for header change

### LINKED ISSUE #16: Sleep for 5 secs while checking for header change in `vote.go`
Continuously pinging the blockchain for new block without any break isn't a good way of doing things.

Yashk767:
resolve #16 

- Added a sleep function of 5 seconds while checking for header change in cmd/vote.go


ashish10677:
This PR contains changes other than only the issue fix. Those changes are also present in https://github.com/razor-network/razor-go/pull/25.
Create a new PR with only the fix for https://github.com/razor-network/razor-go/issues/16

PR #58, opened 2021-07-14
Optimization for RPC calls in vote.go

### LINKED ISSUE #49: Figure out how to reduce RPC calls
# Rationale

We ping the blockchain too many times during voting just to check whether the correct state has arrived or not. Rather than doing that, it'll be better if we calculate the timing of the next state and ping at that particular moment.

# Implementation

Since we know the state length, epoch length, and approximate time to mine a block on a particular blockchain, we can calculate the approximate time when the next state will occur.

ashish10677:
Currently, we fetch a block every 3 secs while voting. This increases the number of RPC calls and is not feasible. Instead of doing that, this fix aims at fetching the block once in every state.  This will reduce the number of RPC calls drastically and optimise the voting process.

Closes #49

### DIFF
--- a/cmd/vote.go
+++ b/cmd/vote.go
@@ -111,8 +111,7 @@ func handleBlock(client *ethclient.Client, account types.Account, blockNumber *b
 			break
 		}
 		_committedData = data
-		time.Sleep(time.Duration(core.StateLength * 2) * time.Second)
-
+		utils.WaitTillNextState()
 	case 1:
 		lastReveal := staker.EpochLastRevealed
 		if _committedData == nil || (lastReveal != nil && lastReveal.Cmp(epoch) >= 0) {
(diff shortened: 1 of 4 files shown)


hrishikeshio:
Once in every state is too risky. What if block time is slightly reduced/increased due to unknown reasons? Once every `blocktime` should be reasonable.

ashish10677:
> Once in every state is too risky. What if block time is slightly reduced/increased due to unknown reasons? Once every `blocktime` should be reasonable.

@hrishikeshio to avoid this issue, we are pinging 10 blocks before a new state, this way we minimize the risk.

hrishikeshio:
> > Once in every state is too risky. What if block time is slightly reduced/increased due to unknown reasons? Once every `blocktime` should be reasonable.
> 
> @hrishikeshio to avoid this issue, we are pinging 10 blocks before a new state, this way we minimize the risk.

still not worth the risk. pinging once every 1 block doesn't sound unreasonable.

ashish10677:
> > > Once in every state is too risky. What if block time is slightly reduced/increased due to unknown reasons? Once every `blocktime` should be reasonable.
> > 
> > 
> > @hrishikeshio to avoid this issue, we are pinging 10 blocks before a new state, this way we minimize the risk.
> 
> still not worth the risk. pinging once every 1 block doesn't sound unreasonable.
> can we make it a configurable parameter?

The only problem in pinging every block is that there are too many requests made, and that might go beyond the Infura limit as in Matic, a block is mined every 2-3 seconds. So in a minute, there will be 20-30 requests, just to check the state.

SkandaBhat:
Makes sense to have it configurable. I might want to play around with RPC call frequency to optimise my rewards after paying out infura/quicknode etc.

ashish10677:
I'll be sending a new PR with a configurable wait time for the next state.

PR #72, opened 2021-07-23
Added test file for cmd/stake.go

### LINKED ISSUE #69: Test cases for stake in `go-razor`

Yashk767:
resolve #69 
- Test file created for cmd/stake.go
- Includes tests for approve and stakeCoins function

Setting Up Testing Environment:
- Start ganache-cli with a menmonic
`ganache-cli -a 15 -b 2 -d -m "slide margin library genuine elbow enhance acoustic genre install wing angry click"
`
- Deploy the contracts on ganache
`npm run deploy:ganache
`
- Get the faucets in accounts from which transactions would be done.
- Replace the deployed addresses of contracts in razor-go/core/constants.go with the new deployed addresses on ganache. 
- Run `npm run build-all` to build razor-go and get the bindings.
- Provide your address for testing in cmd/stake_test.go and start testing functions by running the test files.

### DIFF
--- a/cmd/stake.go
+++ b/cmd/stake.go
@@ -52,7 +52,7 @@ var stakeCmd = &cobra.Command{
 	},
 }
 
-func approve(txnArgs types.TransactionOptions) {
+func approve(txnArgs types.TransactionOptions) int {
 	tokenManager := utils.GetTokenManager(txnArgs.Client)
 	opts := utils.GetOptions(false, txnArgs.AccountAddress, "")
 	allowance, err := tokenManager.Allowance(&opts, common.HexToAddress(txnArgs.AccountAddress), common.HexToAddress(core.StakeManagerAddress))
@@ -61,6 +61,7 @@ func approve(txnArgs types.TransactionOptions) {
 	}
 	if allowance.Cmp(txnArgs.Amount) >= 0 {
 		log.Info("Sufficient allowance, no need to increase")
+		return 1
 	} else {
 		log.Info("Sending Approve transaction...")
 		txnOpts := utils.GetTxnOpts(txnArgs)
@@ -70,7 +71,8 @@ func approve(txnArgs types.TransactionOptions) {
 		}
 		log.Info("Approve transaction sent...\nTxn Hash: ", txn.Hash())
 		log.Info("Waiting for transaction to be mined....")
-		utils.WaitForBlockCompletion(txnArgs.Client, fmt.Sprintf("%s", txn.Hash()))
+		approvedStatus := utils.WaitForBlockCompletion(txnArgs.Client, fmt.Sprintf("%s", txn.Hash()))
+		return approvedStatus
 	}
 }
 
--- a/cmd/stake.go
+++ b/cmd/stake.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"razor/core"
 	"razor/core/types"
@@ -47,12 +48,14 @@ var stakeCmd = &cobra.Command{
 			ChainId:        core.ChainId,
 			GasMultiplier:  config.GasMultiplier,
 		}
-		approve(txnArgs)
-		stakeCoins(txnArgs)
+		approveErr := approve(txnArgs)
+		utils.CheckError("Approve error: ", approveErr)
+		stakeErr := stakeCoins(txnArgs)
+		utils.CheckError("Stake error: ", stakeErr)
 	},
 }
 
-func approve(txnArgs types.TransactionOptions) int {
+func approve(txnArgs types.TransactionOptions) error {
 	tokenManager := utils.GetTokenManager(txnArgs.Client)
 	opts := utils.GetOptions(false, txnArgs.AccountAddress, "")
 	allowance, err := tokenManager.Allowance(&opts, common.HexToAddress(txnArgs.AccountAddress), common.HexToAddress(core.StakeManagerAddress))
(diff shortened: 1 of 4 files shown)


ashish10677:
@Yashk767 code formatting is not done, build is failing, please format your code using `gofmt` and push.

cmd/stake.go
```diff
@@ -70,7 +71,8 @@ func approve(txnArgs types.TransactionOptions) {
 		}
 		log.Info("Approve transaction sent...\nTxn Hash: ", txn.Hash())
 		log.Info("Waiting for transaction to be mined....")
-		utils.WaitForBlockCompletion(txnArgs.Client, fmt.Sprintf("%s", txn.Hash()))
+		approvedStatus := utils.WaitForBlockCompletion(txnArgs.Client, fmt.Sprintf("%s", txn.Hash()))
+		return approvedStatus
```
ashish10677, line 75: This can be changed to `return utils.WaitForBlockCompletion(...)`

cmd/stake_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"math/big"
+	"razor/core"
+	"razor/core/types"
+	"razor/utils"
+	"testing"
+)
+
+func getTransactionArgs() types.TransactionOptions{
+	password := "12345"
+	address := "0x53baCf1E1C6E14BB5700178Acf426dF6f0d5A949" // Provide your address for testing
```
ashish10677, line 16: Instead of hardcoding address, can't we generate address dynamically every time as we do in the `create` command?

cmd/stake_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	type args struct {
+		txnOpts types.TransactionOptions
+	}
+	tests := []struct {
+		name string
+		args args
+	}{
+		{
+			name: "Test1: Staker is able to stake for the first time.",
```
ashish10677, line 80: What about tests for staking below minimum stake?

ashish10677 (changes requested):

cmd/stake_test.go
Yashk767: If I generate the addresses dynamically in the same file, I also need to seed some SC's to the address which is generated just now.So I have to make that transfer work. I am working on it.

cmd/stake_test.go
Yashk767: Yes, I will add it. Should I check the condition in the go function itself because that would be an instant check on running go function. If we are not checking that than by default it would check from bindings Stake function where it would return an revert error of Minimum stake when it hit the commit stake after time and run stake.

ashish10677: We need to check from go functions, not the bindings. The purpose of writing these tests are that our go functions are tested properly, bindings don't need to be tested.

cmd/stake_test.go
```diff
@@ -23,7 +24,7 @@ func getTransactionArgs() types.TransactionOptions{
 	if err != nil {
 		log.Fatal(err)
 	}
-	balance, err := utils.FetchBalance(client, address)
+	balance, _ := utils.FetchBalance(client, address)
```
ashish10677, line 27: Why is the error not being caught over here?

cmd/stake_test.go
```diff
@@ -37,19 +38,45 @@ func getTransactionArgs() types.TransactionOptions{
 	}
 	return txnArgs
 }
+func getTransactionArgs1() types.TransactionOptions {
```
ashish10677, line 41: Please rename this function, `getTransactionArgs1` is not a good name at all.

cmd/stake_test.go
```diff
@@ -91,14 +118,16 @@ func Test_stakeCoins(t *testing.T) {
...
-				approve(txnArgs)
-				stakeCoins(txnArgs)
+				approveErr := approve(txnArgs)
+				utils.CheckError("Approve error: ", approveErr)
+				stakeErr := stakeCoins(txnArgs)
+				utils.CheckError("Stake error: ", stakeErr)
 
-				newStakerID,_ := stakeManager.GetStakerId(&opts, common.HexToAddress(txnArgs.AccountAddress))
-				staker,_ := utils.GetStaker(txnArgs.Client, txnArgs.AccountAddress, newStakerID)
+				newStakerID, _ := stakeManager.GetStakerId(&opts, common.HexToAddress(txnArgs.AccountAddress))
```
ashish10677, line 129: Add condition to check error here

cmd/stake_test.go
```diff
@@ -91,14 +118,16 @@ func Test_stakeCoins(t *testing.T) {
...
+				staker, _ := utils.GetStaker(txnArgs.Client, txnArgs.AccountAddress, newStakerID)
```
ashish10677, line 130: Here as well

ashish10677 (changes requested):

SkandaBhat:
How do I setup tests? This PR needs modifications in the readme as well. @Yashk767

Yashk767:
> How do I setup tests? This PR needs modifications in the readme as well. @Yashk767

@SkandaBhat As we are trying to make the tests dynamic by generating the random addresses and seed them with faucets in the same file and than performing transactions with them but it is not being done yet. Currently we are asking the user to change the hard-coded address and replace it with its own. So I added setting up the testing environment in the PR info as of now, should we add it in readme after making it dynamic if its okay.

SkandaBhat:
Closing this as we are employing a different approach. cc: @ashish10677 @Yashk767

PR #74, opened 2021-07-26
Add wait time as a configurable parameter.

### LINKED ISSUE #49: Figure out how to reduce RPC calls

ashish10677:
Users can now set `wait` as a configurable parameter that sets the time for which voting is put to sleep. This will help optimize the number of RPC calls.

Closes #49

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ RAZOR = ${BIN_DIR}/razor
 
 all: fetch_bindings install_razor set_config
 build: install_razor set_config
+docker-build: fetch_bindings install_razor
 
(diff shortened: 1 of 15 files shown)


SkandaBhat:
related to #58

Issue #57, opened 2021-07-13
Dockerise razor-go


SkandaBhat:
# Rationale
There are lot of environment specific nuances that could surface problems with razor-go operation.

# Implementation
Dockerise the node so that we have control over dependencies and environment.

PR #85, opened 2021-07-28
dockerize razor-go

OxHimanshu:
#57

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,8 +18,7 @@ RUN apt-get update && \
     apt-get install -y  wget && \
     apt-get install -y build-essential && \
     apt-get -y install solc && \
-    apt install -y protobuf-compiler && \
-#RUN    useradd -m docker && echo "docker:docker" | chpasswd && adduser docker sudo 
+    apt install -y protobuf-compiler && \ 
     curl -sL https://deb.nodesource.com/setup_12.x -o nodesource_setup.sh && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+    apt-get install -y nodejs && \
+    npm install --silent && \
+    mkdir tmp/ && cd tmp/ && wget https://golang.org/dl/go1.16.6.linux-amd64.tar.gz && \
+    tar -xvf go1.16.6.linux-amd64.tar.gz && \
+    mv -f go ../../../usr/local && pwd && \
+    cd ../ && rm -rf tmp && \
+    go version && go get -u github.com/ethereum/go-ethereum && \
+    cd ../../go/pkg/mod/github.com/ethereum/go-ethereum@v1.10.6 && chmod +777 build &&  make && make devtools
+RUN npm run
+#RUN ./razor setconfig --provider https://infura/v3/matic --gasmultiplier 1.5
```
ashish10677, line 35: @SkandaBhat should we run setconfig in the docker image or not?

SkandaBhat: Ya we should be able to. This is a user facing feature, and they should have to be able to set config.

ashish10677: So this line should be uncommented right?

SkandaBhat: yes

PR #94, opened 2021-07-30
Changed functionality of commission to commission rate

### LINKED ISSUE #68: Set Delegation Commission sets commission * 10**18 instead of input
#### Describe the bug
Currently in setDelegation.go , commission that we input is considered as the amount which staker would charge for the delegators to delegate. But in actual, commission is set as a commission percent rate in razor contracts for delegation. 

#### Expected behaviour
eg .if the commission is 5 , than 5 percent of the staked s amount would be considered as a commission when delegator withdraws

#### Actual behaviour
If the input is 5 , corresponding amount in wei is considered as a commission

Yashk767:
resolve #68 

- Changed the functionality of staker's commission from commission amount(amount in wei) to commission rate.

### DIFF
--- a/cmd/setDelegation.go
+++ b/cmd/setDelegation.go
@@ -62,7 +62,6 @@ to quickly create a Cobra application.`,
 		}
 
 		_commission, ok := new(big.Int).SetString(commission, 10)
-		commissionAmountInWei := big.NewInt(1).Mul(_commission, big.NewInt(1e18))
 		if !ok {
 			log.Fatal("Set string: error")
 		}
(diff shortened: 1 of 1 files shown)


cmd/setDelegation.go
```diff
@@ -73,37 +72,37 @@ to quickly create a Cobra application.`,
...
 		}
 
 	},
 }
 
-func SetCommission(client *ethclient.Client, stakeManager *bindings.StakeManager, stakerId *big.Int, txnOpts *bind.TransactOpts, commissionAmountInWei *big.Int) {
-	log.Infof("Setting the commission value of Staker %s to %s", stakerId, commissionAmountInWei)
-	commissionTxn, err := stakeManager.SetCommission(txnOpts, commissionAmountInWei)
+func SetCommission(client *ethclient.Client, stakeManager *bindings.StakeManager, stakerId *big.Int, txnOpts *bind.TransactOpts, commission *big.Int) {
+	log.Infof("Setting the commission value of Staker %s to %s", stakerId, commission)
```
ashish10677, line 88: Add a `%` symbol here. It should read `Setting the commission value of Staker 1 to 10%`, something like this.

ashish10677 (changes requested):

cmd/setDelegation.go
```diff
@@ -85,7 +85,7 @@ to quickly create a Cobra application.`,
 }
 
 func SetCommission(client *ethclient.Client, stakeManager *bindings.StakeManager, stakerId *big.Int, txnOpts *bind.TransactOpts, commission *big.Int) {
-	log.Infof("Setting the commission value of Staker %s to %s", stakerId, commission)
+	log.Infof("Setting the commission value of Staker %s to %s %%", stakerId, commission)
```
ashish10677, line 88: Can you remove the space after %s, 
```
log.Infof("Setting the commission value of Staker %s to %s%%", stakerId, commission)
```
This is better

cmd/setDelegation.go
```diff
@@ -94,7 +94,7 @@ func SetCommission(client *ethclient.Client, stakeManager *bindings.StakeManager
 }
 
 func DecreaseCommission(client *ethclient.Client, stakeManager *bindings.StakeManager, stakerId *big.Int, txnOpts *bind.TransactOpts, commission *big.Int) {
-	log.Infof("Decreasing the commission value of Staker %s to %s", stakerId, commission)
+	log.Infof("Decreasing the commission value of Staker %s to %s %%", stakerId, commission)
```
ashish10677, line 97: Here as well.

ashish10677 (changes requested):

ashish10677 (approved):

PR #114, opened 2021-08-10
List all available accounts

### LINKED ISSUE #110: List all accounts that are accessible by the client
# Rationale

We need a feature to list all the accounts whose Keystore file is present in the `.razor` folder. This feature is required because the user sometimes might not know which account does he/she has access to and which he/she needs to import.

# Implementation

There is a function, `keystore.Accounts` that lists all the accounts. Just display the account addresses of those accounts. Check [this](https://github.com/razor-network/razor-go/blob/69e7d6b5e4f70bbe330cc6ec952baa7abb1e35a4/accounts/accounts.go#L37) for more details.

Yashk767:
Resolve #110 
- listAccounts command created in order to display all the available accounts

### DIFF
--- a/cmd/listAccounts.go
+++ b/cmd/listAccounts.go
@@ -1,0 +1,28 @@
+package cmd
+
+import (
+	"github.com/ethereum/go-ethereum/accounts/keystore"
+	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


ashish10677 (approved):
LGTM

PR #115, opened 2021-08-10
Made gas price a configurable parameter.

### LINKED ISSUE #105: Set gas price as a configurable parameter
Gas Price is predicted by calling the `SuggestGasPrice` method. But this value is too high right now. We can let the user choose the gas price for the time being and if he/she doesn't choose a value, then the `SuggestGasPrice` method will be used to predict the gas price.

ashish10677:
Closes #105


utils/options.go
```diff
@@ -40,11 +40,14 @@ func GetTxnOpts(transactionData types.TransactionOptions) *bind.TransactOpts {
...
 }
 
-func getGasPrice(client *ethclient.Client, gasMultiplier float32) *big.Int {
+func getGasPrice(client *ethclient.Client, config types.Configurations) *big.Int {
 	gas, err := client.SuggestGasPrice(context.Background())
 	if err != nil {
 		log.Fatal(err)
 	}
-	gasPrice := MultiplyFloatAndBigInt(gas, float64(gasMultiplier))
+	if config.GasPrice != 0 {
```
SkandaBhat, line 48: Can be more optimal:

if gasPrice == 0 gas = suggestGasPrice
else gas = gasPrice

Currently we are calling SuggestGasPrice every time.

SkandaBhat (changes requested):

utils/options.go
ashish10677: Done

PR #122, opened 2021-08-11
v0.1.5 support

### LINKED ISSUE #103: Support `v0.1.5` version of contracts.

ashish10677:
The `v0.1.5` version of the smart contract doesn't use block hashes while finding an iteration to propose the block. Instead VoteManager.RandaoHash is used. This change is a breaking change in the client and it has been implemented here.

The file `random.go` is deleted because there is no smart contract function of random.sol that is being used in the client now.
Closes #103


SkandaBhat (approved):
lgtm

PR #139, opened 2021-08-18
Propose only when iteration is less than last block in alt blocks when no. of proposed blocks == max alt blocks

### LINKED ISSUE #95: Dont propose if your iteration < last block in alt blocks if num proposed blocks = max alt blocks
# Rationale
Dont know if this issue exists please verify

we have maxaltblocks parameter in contracts which is currently set at 5

lets say 5 blocks have already been proposed with following iterations
[1,12,100,1000,1232]

now current stakers iteration is 2000. The is no point proposing since the block will not be included in proposed blocks anyway since contract only stores `maxaltblocks` numebr of blocks.

in this case, dont propose to save on gas fees.

Yashk767:
fix #95


cmd/propose.go
```diff
@@ -59,7 +59,30 @@ func Propose(client *ethclient.Client, account types.Account, config types.Confi
...
+	numOfProposedBlocks, err := utils.GetNumberOfProposedBlocks(client, account.Address, epoch)
+	if err != nil {
+		log.Error(err)
+	}
+	maxAltBlocks, err := utils.GetMaxAltBlocks(client, account.Address)
+	if err != nil {
+		log.Error(err)
+	}
+	if numOfProposedBlocks.Cmp(maxAltBlocks) == 0 || numOfProposedBlocks.Cmp(maxAltBlocks) == 1 {
+		log.Infof("Number of blocks proposed: %s, which is equal or greater than maximum alternative blocks allowed", numOfProposedBlocks)
```
SkandaBhat, line 71: What is Infof?

ashish10677: `Infof` is like `printf`, a statement that supports string formatting.

cmd/propose.go
```diff
@@ -59,7 +59,30 @@ func Propose(client *ethclient.Client, account types.Account, config types.Confi
...
-
+	numOfProposedBlocks, err := utils.GetNumberOfProposedBlocks(client, account.Address, epoch)
+	if err != nil {
+		log.Error(err)
+	}
+	maxAltBlocks, err := utils.GetMaxAltBlocks(client, account.Address)
+	if err != nil {
+		log.Error(err)
+	}
+	if numOfProposedBlocks.Cmp(maxAltBlocks) == 0 || numOfProposedBlocks.Cmp(maxAltBlocks) == 1 {
```
ashish10677, line 70: Can't you check for >= 0 directly? Instead of checking 2 conditions?

cmd/propose.go
```diff
@@ -59,7 +59,30 @@ func Propose(client *ethclient.Client, account types.Account, config types.Confi
...
+		log.Infof("Number of blocks proposed: %s, which is equal or greater than maximum alternative blocks allowed", numOfProposedBlocks)
+		log.Info("Comparing  iterations...")
+		lastBlockIndex := big.NewInt(0)
+		lastBlockIndex.Sub(numOfProposedBlocks, big.NewInt(1))
```
ashish10677, line 74: This and the above statement can be consolidated into one:
```
lastBlockIndex := big.NewInt(0).Sub(numOfProposedBlocks, big.NewInt(1))
```

cmd/propose.go
```diff
@@ -59,7 +59,30 @@ func Propose(client *ethclient.Client, account types.Account, config types.Confi
...
+		lastProposedBlockStruct, err := utils.GetProposedBlock(client, account.Address, epoch, lastBlockIndex)
+		if err != nil {
+			log.Error(err)
+		}
+		lastIteration := lastProposedBlockStruct.Block.Iteration
+		if lastIteration.Cmp(big.NewInt(int64(iteration))) == -1 {
```
ashish10677, line 80: This is not wrong, but I feel, writing `<0` makes it easier to understand.

ashish10677 (changes requested):

cmd/propose.go
Yashk767: yes, I can do that.

ashish10677 (approved):

PR #140, opened 2021-08-18
Ux improvements for incentivized test-net.

### LINKED ISSUE #136: Modifications/Improvements for Incentivised testnet
- [x] There is cobra cli placeholder text displayed while running command
- [x] running razor without any flags/commands should run help command
- [x] description for all commands needs to be looked at
- [x] —address needs a shorthand flag. —amount needs to be changed to —value and -v
- [x] Address needs to be validated before sending command (throw if address not in list)
- [x] If wrong password, fail first
- [x] setconfig should be in camelcase
- [x] Waiting for x blocks - should be waiting for x seconds
- [x] Should not allow withdraw if not unstaked
- [x] dockerise-build to be renamed to build-noargs
- [x] add a warning if not https

ashish10677:
Closes #136


SkandaBhat (approved):
Looks good to me.

PR #151, opened 2021-08-20
Feature password flag

### LINKED ISSUE #131: Read password from file
# Rationale

In servers, we need a way to read account password from a file to make it easier to run vote in an automated way.

# Implementation

Add an argument --password and supply a text file that razor-go can read from.

eg., ./razor vote --address <addr> --password $HOME/password

Contents of $HOME/password:
password of the respective account in the first line. Anything after newline or EOF will be disregarded

Yashk767:
fix #131 
- Added utility functions GetPasswordFromFile() and IsFlagPassed()
- Added a feature to read the password from file when password flag is provided with a path to file as its value.


cmd/create.go
```diff
@@ -39,4 +46,10 @@ var createCmd = &cobra.Command{
 
 func init() {
 	rootCmd.AddCommand(createCmd)
+
+	var (
+		Password string
+	)
+
+	createCmd.Flags().StringVarP(&Password, "password", "", "", "password of the user")
```
ashish10677, line 54: Change description to "password file path to protect the keystore"

cmd/resetLock.go
```diff
@@ -24,7 +24,14 @@ to quickly create a Cobra application.`,
 		config, err := GetConfigData()
 		utils.CheckError("Error in getting config data: ", err)
 
-		password := utils.PasswordPrompt()
+		var password string
```
ashish10677, line 27: Can this whole block be converted to a function, and used as a one-line call to that function. This piece of code is being used a lot.

ashish10677 (changes requested):

cmd/resetLock.go
Yashk767: yes this could be converted. We can make a function with argument as flagset type, as we need cmd.Flags() to get the flags from the command.

SkandaBhat:
We need a warning if a user uses the --password flag.

utils/password.go
```diff
@@ -45,3 +49,37 @@ func validatePrivateKey(input string) error {
...
+		log.Fatal(err)
+	}
+	return ""
+}
+
+func AssignPassword(flagset *pflag.FlagSet) string {
+	if IsFlagPassed("password") {
+		passwordPath, _ := flagset.GetString("password")
+		prompt := promptui.Prompt{
+			Label:     "Setting password from file at described path, are you sure?",
```
SkandaBhat, line 75: The point of fetching passwords from files is to prevent prompts in CI settings. 

We need to just print a warning if the flag is passed, not as for prompt.

SkandaBhat (approved):

SkandaBhat (changes requested):
Please remove prompt.

utils/password.go
Yashk767: Okay, I will change it.

ashish10677: @Yashk767 you can use `log.Warn` for sending out a warning.

SkandaBhat:
This Branch was force pushed. Please create a new PR and get it reviewed from @ashish10677. Closing this.

PR #158, opened 2021-08-22
Exponential retry implemented while proposing block.

### LINKED ISSUE #156: Exponential retry mechanism needs to be implemented for connecting to RPC URL.
# Rationale

Sometimes, the RPC URL throws an error and kills the vote command. This can be bad for the staker because they can be slashed for inactivity.

To avoid this, an exponential retry mechanism needs to be implemented.

# Implementation

1. Try to connect to RPC URL.
2. If it throws an error, logs the error and try after 2 secs, then 4 secs, then 8 secs and so on...
3. For the sake of how many tries should the client do before completely killing the process, we can add a new config parameter in which the staker can decide how many retries they need.

ashish10677:
Closes #156 

### DIFF
--- a/cmd/vote.go
+++ b/cmd/vote.go
@@ -7,11 +7,13 @@ import (
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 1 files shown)


SkandaBhat (approved):

SkandaBhat:
Looks good to me.

PR #162, opened 2021-08-24
Checked Eth Balance is 0 instead of RZR balance check.

### LINKED ISSUE #157: Ether balance needs to be checked instead of RZR balance in withdraw.go
#### Describe the bug
Currently we are aborting the withdraw process if RZR balance of staker is 0, so if a staker stakes all of his account RZR's than he is not able to withdraw.

#### Expected behaviour
Staker should able to withdraw even when account RZR balance is 0.

Check ether balance instead of RZR balance.

Yashk767:
fixes #157

### DIFF
--- a/cmd/withdraw.go
+++ b/cmd/withdraw.go
@@ -32,17 +32,12 @@ to quickly create a Cobra application.`,
 
 		client := utils.ConnectToClient(config.Provider)
 
-		balance, err := utils.FetchBalance(client, address)
-		utils.CheckError("Error in fetching balance for account"+address+": ", err)
-
-		if balance.Cmp(big.NewInt(0)) == 0 {
-			log.Fatal("Balance is 0. Aborting...")
-		}
-
 		_stakerId, ok := new(big.Int).SetString(stakerId, 10)
 		if !ok {
 			log.Fatal("Set string error in converting staker id")
 		}
+
+		utils.CheckEthBalanceIsZero(client, address)
 
 		checkForCommitStateAndWithdraw(client, types.Account{
 			Address:  address,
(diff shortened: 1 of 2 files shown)


ashish10677 (approved):
LGTM

Yashk767:
Should we check this condition of CheckEthBalanceIsZero in other state dependent functions like delegate and unstake as well ?. As we have to wait till the commit state to hit to know the status, if we check it earlier there is no need to wait till state is being hit.

ashish10677:
> Should we check this condition of CheckEthBalanceIsZero in other state dependent functions like delegate and unstake as well ?. As we have to wait till the commit state to hit to know the status, if we check it earlier there is no need to wait till state is being hit.

Yes, that makes sense. Can you open an issue and raise a PR for those as well?

Yashk767:
> > Should we check this condition of CheckEthBalanceIsZero in other state dependent functions like delegate and unstake as well ?. As we have to wait till the commit state to hit to know the status, if we check it earlier there is no need to wait till state is being hit.
> 
> Yes, that makes sense. Can you open an issue and raise a PR for those as well?

yes

SkandaBhat (changes requested):
> Yes, that makes sense. Can you open an issue and raise a PR for those as well?

Please link the created issue here.

Yashk767:
> > Should we check this condition of CheckEthBalanceIsZero in other state dependent functions like delegate and unstake as well ?. As we have to wait till the commit state to hit to know the status, if we check it earlier there is no need to wait till state is being hit.
> 
> Yes, that makes sense. Can you open an issue and raise a PR for those as well?

I have created the issue #164

ashish10677 (approved):

PR #163, opened 2021-08-24
Modify asset status implemented.

### LINKED ISSUE #148: Support modifyAsset on `razor-go`
We have an option to modifyAsset in the razor contracts. The `razor-go` client must support it as well.

ashish10677:
Closes #148 

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,12 +79,12 @@ $ ./razor stake --address 0x5a0b54d5dc17e0aadc383d2db43b0a0d3e029c4c --amount 10
 
 If you are a staker you can accept delegation from delegators and charge a commission from them.
 ```
-$ ./razor setDelegation --address <address> --status <true_or_false> --commission <commission>
+$ ./razor setDelegation --address <address> --status=<true_or_false> --commission <commission>
 ```
 
 Example:
 ```
-$ ./razor setDelegation --address 0x5a0b54d5dc17e0aadc383d2db43b0a0d3e029c4c --status true --commission 100
+$ ./razor setDelegation --address 0x5a0b54d5dc17e0aadc383d2db43b0a0d3e029c4c --status=true --commission 100
 ```
 
 ### Delegate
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -79,12 +79,12 @@ $ ./razor stake --address 0x5a0b54d5dc17e0aadc383d2db43b0a0d3e029c4c --amount 10
 
 If you are a staker you can accept delegation from delegators and charge a commission from them.
 ```
-$ ./razor setDelegation --address <address> --status <true_or_false> --commission <commission>
+$ ./razor setDelegation --address <address> --status=<true_or_false> --commission <commission>
```
Yashk767, line 82: Yes, this change was needed. Can we have an additional validation like values assigned to status flag must include '=' , because if we are not using '=' than also it is valid but the value is always getting set to true.

eg. ./razor setDelegation --address 0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BD --status false or
./razor setDelegation --address 0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BD --status .

Both are setting delegationAcceptance to true irrespective of the value assigned to status flag without '='.

SkandaBhat: We need to figure out why before we move to another solution.

Adding '=' just for one argument is an anti-pattern and should be avoided.

README.md
```diff
@@ -183,7 +213,7 @@ $  ./razor createJob --address 0x5a0b54d5dc17e0aadc383d2db43b0a0d3e029c4c -n eth
 ```
 
 ### Create Collection
-You can create new collections using `creteCollection` command. This command will work only for admins.
+The admin can create new collections using `creteCollection` command.
```
SkandaBhat, line 216: Better to avoid mentioning the actors directly.

eg., 
### Create Collection
Create new collections using 'createCollection' command. 
Note: This command is restricted to "Admin" Role.

README.md
ashish10677: https://github.com/spf13/cobra/issues/613

This is the issue @SkandaBhat

README.md
ashish10677: Ok, I'll do the change.

README.md
SkandaBhat: Is it possible to read it as a string and convert it to boolean?

`./razor setDelegation --address 0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BD --status "false"`

ashish10677: Yes it is possible

SkandaBhat: Lets do it this way then. Status can be only one of ["true"|"false"]. Anything else will return.

PR #172, opened 2021-08-26
Implement algorithm to support large amount of votes during giveSorted.

### LINKED ISSUE #88: Implement back off algorithm while calling `giveSorted` in dispute
# Rationale

Right now, we call `giveSorted` with 1000 values without considering the fact that the client might surpass the block gas limit while calling this function. This must be avoided.

# Implementation

To avoid this, a binary backoff algorithm must be implemented, where first we estimate gas for the entire votes, and if the value is greater than the block gas limit, we go on decreasing that value by half, till the value is less than the block gas limit.

ashish10677:
Fixes #88

### DIFF
--- a/cmd/dispute.go
+++ b/cmd/dispute.go
@@ -1,11 +1,13 @@
 package cmd
 
 import (
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/ethclient"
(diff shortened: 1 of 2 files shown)


SkandaBhat:
Looks good to me.

ashish10677:
> Looks good to me.

Can you approve this as well @SkandaBhat

SkandaBhat (approved):

PR #180, opened 2021-09-01
Transact fractional RZRs

### LINKED ISSUE #169: Add a feature to stake and unstake/withdraw fractional RZRs.
# Rationale

Right now the amount for stake/unstake/withdraw is assumed to be in RZRs. This makes it impossible to use fractional amount of Rzrs.

# Implementation

A new flag `--pow` can be introduced that'll accept the multiplier for the Razors.

Yashk767:
Closes #169

### DIFF
--- a/cmd/delegate.go
+++ b/cmd/delegate.go
@@ -26,14 +26,15 @@ Example:
 		password := utils.AssignPassword(cmd.Flags())
 		address, _ := cmd.Flags().GetString("address")
 		stakerId, _ := cmd.Flags().GetString("stakerId")
-		value, _ := cmd.Flags().GetString("value")
 
 		client := utils.ConnectToClient(config.Provider)
 
 		balance, err := utils.FetchBalance(client, address)
 		utils.CheckError("Error in fetching balance for account "+address+": ", err)
 
-		valueInWei := utils.GetAmountWithChecks(value, balance)
+		valueInWei := utils.AssignAmountInWei(cmd.Flags())
+		utils.GetAmountWithChecks(valueInWei, balance)
+
 		utils.CheckEthBalanceIsZero(client, address)
 
 		_stakerId, ok := new(big.Int).SetString(stakerId, 10)
@@ -53,7 +54,7 @@ Example:
 
 		approve(txnOpts)
 
-		log.Infof("Delegating %s razors to Staker %s", value, _stakerId)
+		log.Infof("Delegating razors to Staker %s", _stakerId)
 		delegationTxnOpts := utils.GetTxnOpts(txnOpts)
 		epoch, err := WaitForCommitState(client, address, "delegate")
 		utils.CheckError("Error in fetching epoch: ", err)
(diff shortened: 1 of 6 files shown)


utils/math.go
```diff
@@ -66,16 +67,42 @@ func AllZero(bytesValue [32]byte) bool {
...
+	amountInWei := big.NewInt(1).Mul(amount, big.NewInt(1e18))
+	return amountInWei
+}
+
+func GetFractionalAmountInWei(amount *big.Int, power string) *big.Int {
+	_power, err := new(big.Int).SetString(power, 10)
+	if !err {
+		log.Fatal("SetString: error")
+	}
+	amountInWei := big.NewInt(1).Mul(amount, big.NewInt(0).Exp(big.NewInt(10), _power, nil))
```
ashish10677, line 87: Make this as `big.NewInt(1).Exp(big.NewInt(10), _power)`

utils/math_test.go
```diff
@@ -246,7 +246,7 @@ func TestGetAmountWithChecks(t *testing.T) {
 		{
 			name: "Test When amount is equal to balance",
 			args: args{
-				amount:  "1000",
+				amount:  big.NewInt(1).Mul(big.NewInt(1000), big.NewInt(1e18)),
 				balance: big.NewInt(1).Mul(big.NewInt(1000), big.NewInt(1e18)),
 			},
 			want:          big.NewInt(1).Mul(big.NewInt(1000), big.NewInt(1e18)),
```
ashish10677, line 252: Write Test cases for all the new functions introduced in utils/math.go

ashish10677 (changes requested):

ashish10677:
Please link the issues to this PR.
image

utils/math.go
```diff
@@ -66,16 +67,42 @@ func AllZero(bytesValue [32]byte) bool {
 	return true
 }
 
-func GetAmountWithChecks(amount string, balance *big.Int) *big.Int {
+func GetAmountWithChecks(amountInWei *big.Int, balance *big.Int) *big.Int {
+	if amountInWei.Cmp(balance) > 0 {
+		log.Fatal("Not enough balance")
+	}
+	return amountInWei
```
ashish10677, line 74: No need to return a value if you're not using it.

cmd/delegate.go
```diff
@@ -26,14 +26,15 @@ Example:
...
-		value, _ := cmd.Flags().GetString("value")
 
 		client := utils.ConnectToClient(config.Provider)
 
 		balance, err := utils.FetchBalance(client, address)
 		utils.CheckError("Error in fetching balance for account "+address+": ", err)
 
-		valueInWei := utils.GetAmountWithChecks(value, balance)
+		valueInWei := utils.AssignAmountInWei(cmd.Flags())
+		utils.GetAmountWithChecks(valueInWei, balance)
```
ashish10677, line 36: Change the name of this function, to something like `CheckAmountAndBalance`. The `get` word doesn't make much sense here as we are not getting any value here.

cmd/delegate.go
```diff
@@ -53,7 +54,7 @@ Example:
 
 		approve(txnOpts)
 
-		log.Infof("Delegating %s razors to Staker %s", value, _stakerId)
+		log.Infof("Delegating razors to Staker %s", _stakerId)
```
ashish10677, line 57: Can you convert the values in RZR and then display those here. Even if you have to show it in decimals.

ashish10677 (changes requested):

utils/math.go
Yashk767: Actually we are using  the return value while testing the function.
ref - https://github.com/razor-network/razor-go/blob/develop/utils/math_test.go#L264

ashish10677 (approved):

PR #181, opened 2021-09-01
Auto unstake and withdraw if stake < minStake while voting

### LINKED ISSUE #179: Unstake and AutoWithdraw if stake goes below minimum stake.
# Rationale

If the stake goes below the minimum stake while the voting is happening, due to a change In governance parameter, the staker might stop voting due to the conditions in razor-go. And when they resume voting, there is a chance they might incur penalties.

To avoid this whenever the stake is below the minimum, the client should automatically unstake and withdraw funds.

# Implementation

In `vote.go` there is a check for the minimum stake, if the staked value is below minimum stake but greater than 0, call the unstake and withdraw functions.

Yashk767:
Closes #179

### DIFF
--- a/cmd/vote.go
+++ b/cmd/vote.go
@@ -83,7 +83,8 @@ func handleBlock(client *ethclient.Client, account types.Account, blockNumber *b
 		return
 	}
 	if stakerId.Cmp(big.NewInt(0)) == 0 {
-		log.Fatal("Staker doesn't exist")
+		log.Error("Staker doesn't exist")
+		return
 	}
 	stakedAmount, err := utils.GetStake(client, account.Address, stakerId)
 	if err != nil {
@@ -104,11 +105,13 @@ func handleBlock(client *ethclient.Client, account types.Account, blockNumber *b
 	if stakedAmount.Cmp(minStakeAmount) < 0 {
 		log.Error("Stake is below minimum required. Cannot vote.")
 		if stakedAmount.Cmp(big.NewInt(0)) == 0 {
-			log.Fatal("Stopped voting as total stake is already withdrawn.")
+			log.Error("Stopped voting as total stake is already withdrawn.")
+		} else {
+			log.Info("Auto starting Unstake followed by Withdraw")
+			AutoUnstakeAndWithdraw(client, account, stakedAmount, config)
+			log.Error("Stopped voting as total stake is withdrawn now")
 		}
-		log.Info("Auto starting Unstake followed by Withdraw")
-		AutoUnstakeAndWithdraw(client, account, stakedAmount, config)
-		log.Fatal("Stopped voting as total stake is withdrawn now")
+		return
 	}
 
 	staker, err := utils.GetStaker(client, account.Address, stakerId)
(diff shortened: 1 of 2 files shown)


cmd/vote.go
```diff
@@ -100,7 +104,12 @@ func handleBlock(client *ethclient.Client, account types.Account, blockNumber *b
 	log.Info(aurora.Red("🔲 Block:"), aurora.Red(blockNumber), aurora.Yellow("⌛ Epoch:"), aurora.Yellow(epoch), aurora.Green("⏱️ State:"), aurora.Green(state), aurora.Blue("📒:"), aurora.Blue(account.Address), aurora.BrightBlue("👤 Staker ID:"), aurora.BrightBlue(stakerId), aurora.Cyan("💰Stake:"), aurora.Cyan(stakedAmount), aurora.Magenta("Ξ:"), aurora.Magenta(ethBalance))
 	if stakedAmount.Cmp(minStakeAmount) < 0 {
 		log.Error("Stake is below minimum required. Cannot vote.")
-		return
+		if stakedAmount.Cmp(big.NewInt(0)) == 0 {
+			log.Fatal("Stopped voting as total stake is already withdrawn.")
```
ashish10677, line 108: Let this be log.Error for now. We'll see if we want to stop the process or let it continue.

cmd/vote.go
```diff
@@ -224,6 +233,41 @@ func calculateSecret(account types.Account, epoch *big.Int) []byte {
...
+	txnOpts := utils.GetTxnOpts(txnArgs)
+
+	epoch, err := WaitForCommitState(client, account.Address, "unstake")
+	utils.CheckError("Error in fetching epoch: ", err)
+
+	stakerId, err := utils.GetStakerId(client, account.Address)
+	utils.CheckError("Error in getting staker id: ", err)
+
+	log.Info("Unstaking coins")
+	txn, err := stakeManager.Unstake(txnOpts, epoch, stakerId, txnArgs.Amount)
```
ashish10677, line 256: I was thinking, can you create an `unstake` function in` unstake.go` and then call that function from here itself. We can save a lot of code duplication.

ashish10677 (changes requested):

cmd/vote.go
Yashk767: Even if we don't stop the execution, it will be stopped when it will try to unstake again by throwing error of non positive stake. So should we keep the same or change it to log.Error?

ashish10677:
Link the issue here as well.

cmd/vote.go
ashish10677: Change everywhere to log.Error in vote.go, and return after logging error. Don't stop it by running log.Fatal.

cmd/vote.go
Yashk767: Yeah I was thinking to do that but the issue was related to voting so I didn't change there. So should I create the separate unstake function in the same issue PR ?

cmd/vote.go
Yashk767: Okay, I will change it.

cmd/vote.go
ashish10677: Yeah you can do that, since it is not any issue but only code reformatting.

ashish10677 (approved):

PR #185, opened 2021-09-08
Create test

SkandaBhat:
tldr;

1. Code must be refactored into easy to test functions.
2. Create and utilise interfaces when calling methods from external packages
3. Mock method calls that are not necessary for testing.

### DIFF
--- a/cmd/create_test.go
+++ b/cmd/create_test.go
@@ -1,0 +1,63 @@
+package cmd
+
+import (
+	"testing"
+
+	"github.com/ethereum/go-ethereum/accounts"
+	"github.com/ethereum/go-ethereum/common"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/pflag"
+	"github.com/stretchr/testify/assert"
+)
+
+var AssignPasswordMock func(flagset *pflag.FlagSet) string
+
+var GetDefaultPathMock func() string
+
+var CreateAccountMock func(string, string) accounts.Account
+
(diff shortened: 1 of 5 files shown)


cmd/create_test.go
```diff
@@ -0,0 +1,63 @@
...
+			Address: common.HexToAddress("0x000000000000000000000000000000000000dea1"),
+			URL:     accounts.URL{Scheme: "TestKeyScheme", Path: "test/key/path"},
+		}
+	}
+
+	account := Create(flags, args, razorUtils, razorAccounts)
+	log.Info("Account address: ", account.Address)
+	log.Info("Keystore Path: ", account.URL)
+
+	assert.Equal(t, account.Address, common.HexToAddress("0x000000000000000000000000000000000000dead"), "Created Address should match the supplied Address.")
```
ashish10677, line 62: Does this test pass? Because here address is `"0x000000000000000000000000000000000000dead"` and the `CreateAccountMock` returns `"0x000000000000000000000000000000000000dea1"` which are not equal due to the last character

ashish10677: Indeed the test doesn't pass on CircleCI as well. You need to change the addresses to match.

SkandaBhat: This does not pass. This is for illustration. Check CirlceCI for more details

SkandaBhat:
Closing this and reopening as draft.

PR #232, opened 2021-09-21
Test delegate

### LINKED ISSUE #199: Tests for delegate command

Yashk767:
Closes #199

### DIFF
--- a/cmd/delegate.go
+++ b/cmd/delegate.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"github.com/ethereum/go-ethereum/common"
 	"razor/core"
 	"razor/core/types"
 	"razor/utils"
@@ -35,8 +36,7 @@ Example:
 
 		utils.CheckEthBalanceIsZero(client, address)
 
-		stakeManager := utils.GetStakeManager(client)
-		txnOpts := types.TransactionOptions{
+		txnArgs := types.TransactionOptions{
 			Client:         client,
 			Password:       password,
 			Amount:         valueInWei,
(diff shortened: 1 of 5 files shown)


ashish10677:
Shouldn't flag testing be done here as well? What do you think @Yashk767 ?

Yashk767:
> Shouldn't flag testing be done here as well? What do you think @Yashk767 ?

Actually we are calling approve function before delegate , so here we need the flag inputs for txnArgs in both approve and delegate as well.

Yashk767:
> Shouldn't flag testing be done here as well? What do you think @Yashk767 ?

We can do one thing, we can include approve function call inside delegate and stakeCoins function along with flags and create a CmdInterface and implement functions like approve in it, as it is being used as dependent function so that we can call it from there and get the flags tested as well

ashish10677:
> > Shouldn't flag testing be done here as well? What do you think @Yashk767 ?
> 
> Actually we are calling approve function before delegate , so here we need the flag inputs for txnArgs in both approve and delegate as well.

Ok, so in approve the flags are being tested right?

Yashk767:
> > > Shouldn't flag testing be done here as well? What do you think @Yashk767 ?
> > 
> > 
> > Actually we are calling approve function before delegate , so here we need the flag inputs for txnArgs in both approve and delegate as well.
> 
> Ok, so in approve the flags are being tested right?

At current they are not,  but we can include flags in approve itself and test them there, and extra flag like stakerId in delegate can be included in delegate function itself.  We need to implement flag functions and use them accordingly which I have done for other PR so can we modify approve and do this in some other PR ?

ashish10677:
> At current they are not, but we can include flags in approve itself and test them there, and extra flag like stakerId in delegate can be included in delegate function itself. We need to implement flag functions and use them accordingly which I have done for other PR so can we modify approve later and do this in some other PR ?

Yes

ashish10677:
Also fix the conflicts @Yashk767

ashish10677 (approved):

PR #264, opened 2021-10-05
Maintain list of assetIds for which giveSorted has been called already.

### LINKED ISSUE #237: Maintain assetIds for which giveSorted has been already called.
# Rationale

If giveSorted has been called for a particular `assetId` then we don't need to call it again. But if assetId is different, then `giveSorted` must be called again.

# Implementation

ashish10677:
# Description

`GiveSorted` must be called once for a particular asset id. We now maintain a list of asset ids for which `GiveSorted` has been called, and if the current asset id is not present in that array, only then, `GiveSorted` is called.

Fixes #237 

## Type of change

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/cmd/dispute.go
+++ b/cmd/dispute.go
@@ -10,6 +10,8 @@ import (
 	"github.com/ethereum/go-ethereum/ethclient"
 )
 
+var giveSortedAssetIds []int
+
 func HandleDispute(client *ethclient.Client, config types.Configurations, account types.Account, epoch uint32) {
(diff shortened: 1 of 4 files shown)


SkandaBhat (approved):
LGTM

PR #381, opened 2021-11-30
Added claimBounty command 

### LINKED ISSUE #377: Create command to claim bounty

Yashk767:
# Description

Added claimBounty command to claim bounty for bounty hunter

Fixes #377

## Type of change

Please delete options that are not relevant.

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/cmd/claimBounty.go
+++ b/cmd/claimBounty.go
@@ -56,9 +56,11 @@ func (utilsStruct UtilsStruct) executeClaimBounty(flagSet *pflag.FlagSet) {
 	}
 
 	txn, err := utilsStruct.cmdUtils.claimBounty(config, client, redeemBountyInput, utilsStruct)
-
 	utils.CheckError("ClaimBounty error: ", err)
-	utilsStruct.razorUtils.WaitForBlockCompletion(client, txn.String())
+
+	if txn != core.NilHash {
+		utilsStruct.razorUtils.WaitForBlockCompletion(client, txn.String())
+	}
 }
 
 func claimBounty(config types.Configurations, client *ethclient.Client, redeemBountyInput types.RedeemBountyInput, utilsStruct UtilsStruct) (common.Hash, error) {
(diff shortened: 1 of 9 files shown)


cmd/claimBounty.go
```diff
@@ -0,0 +1,137 @@
...
+		log.Error(err)
+		return core.NilHash, err
+	}
+
+	waitFor := big.NewInt(1).Sub(bountyLock.RedeemAfter, big.NewInt(int64(epoch)))
+	if waitFor.Cmp(big.NewInt(0)) == 1 {
+		log.Debug("Waiting for lock period to get over....")
+
+		//waiting till epoch reaches redeemAfter
+		utilsStruct.razorUtils.Sleep(time.Duration(waitFor.Int64()))
```
ashish10677, line 100: If wait for is 20, so this sleep is for 20 seconds, am I right?

cmd/claimBounty.go
```diff
@@ -0,0 +1,137 @@
...
+	rootCmd.AddCommand(claimBountyCmd)
+	var (
+		Address  string
+		Password string
+		BountyId uint32
+	)
+
+	claimBountyCmd.Flags().StringVarP(&Address, "address", "a", "", "address of the staker")
+	claimBountyCmd.Flags().StringVarP(&Password, "password", "", "", "password path of staker to protect the keystore")
+	claimBountyCmd.Flags().Uint32VarP(&BountyId, "bountyId", "", 0, "bountyId of the bounty hunter")
```
ashish10677, line 130: Can we put b for bountyId? Are we using `b` anywhere else?

ashish10677 (changes requested):

cmd/claimBounty.go
Yashk767: Ohh yes it is, I need it to multiply it by epoch length , and correct it to Sleep(time.Duration(waitFor.Int64()* epochLength)) correct ?

cmd/claimBounty.go
Yashk767: Yes we are using it for bufferPercent.

cmd/claimBounty.go
ashish10677: Yeah, but we also need block time. Can we write a function to calculate the time taken to mine a block by calculating the difference of block time between 2 latest blocks?
Then sleep duration will be `waitFor*epochLength*blockTime`.

We can get the latest block's time from:
https://github.com/ethereum/go-ethereum/blob/163f1665dd5074f64fd4b54360306837c6dee43d/core/types/block.go#L81

cmd/claimBounty.go
ashish10677: Okay cool. No need then.

cmd/claimBounty.go
Yashk767: Yes we can do that

cmd/claimBounty.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	if bountyLock.Amount.Cmp(big.NewInt(0)) == 0 {
+		err = errors.New("bounty amount is 0")
+		log.Error(err)
+		return core.NilHash, err
+	}
+
+	waitFor := big.NewInt(1).Sub(bountyLock.RedeemAfter, big.NewInt(int64(epoch)))
+	if waitFor.Cmp(big.NewInt(0)) == 1 {
```
ashish10677, line 96: I didn't understand this if statement. You are checking if the value of `waitFor == 0` then you are waiting???

cmd/claimBounty.go
ashish10677: Also, even after doing that, we might not have reached the exact `RedeemAfter` period. Put a retry as well.

cmd/claimBounty.go
Yashk767: waitFor is difference in redeemAfter epoch and current epoch.
If waitFor  > 0 than Cmp statement would return 1. So if waitFor is > 0 that means we need to waitFor some time, therefore we are waiting for that specific amount of epochs.

Yashk767: If waitFor == 0 , than we don't need to wait so we are directly calling redeemBlock function without waiting as we would already reach the redeemAfter epoch.

ashish10677: Okay got it.

cmd/claimBounty.go
ashish10677: 2 retry for 1 epoch each.

Yashk767: okay

ashish10677:
@Yashk767 can you update README as well?

ashish10677 (approved):
LGTM

PR #383, opened 2021-12-01
Test create

### LINKED ISSUE #371: Add tests for create.go

Yashk767:
# Description

Increased coverage for cmd/create.go

Fixes #371

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -17,20 +17,26 @@ var createCmd = &cobra.Command{
 
 Example: 
   ./razor create`,
-	Run: func(cmd *cobra.Command, args []string) {
-		utilsStruct := UtilsStruct{
-			razorUtils:   razorUtils,
-			accountUtils: accountUtils,
-		}
-		account, err := utilsStruct.Create(cmd.Flags())
-		utils.CheckError("Create error: ", err)
-		log.Info("Account address: ", account.Address)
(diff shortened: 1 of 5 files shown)


ashish10677:
@Yashk767 there seem to be conflicts

cmd/create.go
```diff
@@ -17,20 +17,27 @@ var createCmd = &cobra.Command{
...
+
+func (utilsStruct UtilsStruct) executeCreate(flagSet *pflag.FlagSet) {
 	password := utilsStruct.razorUtils.AssignPassword(flagSet)
+	account, err := utilsStruct.cmdUtils.Create(password, utilsStruct)
+	utils.CheckError("Create error: ", err)
+	log.Info("Account address: ", account.Address)
+	log.Info("Keystore Path: ", account.URL)
+}
+
+func Create(password string, utilsStruct UtilsStruct) (accounts.Account, error) {
```
ashish10677, line 40: We cannot have `utilsStruct.Create`?

Yashk767: No , as Create function is used in executeCreate() so we need to mock it there. So If we want to mock it than we cannot call it by `UtilsStruct.Create` as we would already be calling it by `utilsStruct.cmdUtils.Create`

ashish10677 (approved):

PR #385, opened 2021-12-02
Options fix

### LINKED ISSUE #382: Remove usage of address from `GetOptions`
# Rationale

`GetOptions` fetches `bind.CallOpts` required for getter methods. This doesn't require the address field.

# Implementation

Remove address field from `GetOptions` and change the code wherever we are calling `GetOptions`.

ashish10677:
# Description

Address is not required while calling getter methods. We have removed it in this PR.

Fixes #382


SkandaBhat (approved):
lgtm

PR #403, opened 2021-12-07
Coverage of propose.go made 100%

### LINKED ISSUE #192: Tests for propose

ashish10677:
Fixes #192

### DIFF
--- a/cmd/propose.go
+++ b/cmd/propose.go
@@ -48,7 +48,7 @@ func (utilsStruct UtilsStruct) Propose(client *ethclient.Client, account types.A
 	log.Debug("Biggest Influence Id: ", biggestInfluenceId)
 	log.Debugf("Biggest influence: %s, Stake: %s, Staker Id: %d, Number of Stakers: %d, Randao Hash: %s", biggestInfluence, staker.Stake, stakerId, numStakers, hex.EncodeToString(randaoHash[:]))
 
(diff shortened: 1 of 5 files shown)


cmd/propose.go
```diff
@@ -140,8 +140,8 @@ func getBiggestInfluenceAndId(client *ethclient.Client, address string, epoch ui
 	return biggestInfluence, biggestInfluenceId, nil
 }
 
-func getIteration(client *ethclient.Client, address string, proposer types.ElectedProposer, utilsStruct UtilsStruct) int {
-	for i := 0; i < 10000000000; i++ {
+func getIteration(client *ethclient.Client, proposer types.ElectedProposer, utilsStruct UtilsStruct) int {
+	for i := 0; i < 10000000; i++ {
```
SkandaBhat, line 144: Why was this number changed?

SkandaBhat (changes requested):

cmd/propose.go
ashish10677: 10000000000 is a very big number, and if the iteration goes almost till that number, it is likely that your block will not be included in the maxAltBlocks. This also slowed down testing if we want to go to the last return after the `for` loop. Hence, I shortened the value.

SkandaBhat: Needs to be accounted for in a separate issue to make sure we document it

ashish10677:
Will open a new PR wrt change in the iteration max value.

PR #502, opened 2022-01-20
Staker cannot commit when one collection is inactive fix

### LINKED ISSUE #478: Staker cannot commit even when any one collection becomes inactive
#### System information

- **OS & Version**: Windows/Linux/OSX
- **Go version:** 
- **razor-go version:** 

#### Backtrace
Screenshot from 2022-01-13 13-29-03
Screenshot from 2022-01-13 13-28-35

shekhar2807:
-Staker can now commit when one of the collection is inactive .

fixes #478 

P.S. - Quick fix tests would be covered after the mockery is done for utils/asset.go in #394


SkandaBhat:
This needs a test case to verify that the fix works.

PR #569, opened 2022-02-14
Test giveSorted

### LINKED ISSUE #193: Tests for dispute

Yashk767:
# Description

Added tests for giveSorted function present in dispute.go

Fixes #193


cmd/dispute_test.go
```diff
@@ -405,3 +405,60 @@ func TestHandleDispute(t *testing.T) {
...
+			args: args{
+				sortedStakers: []uint32{2, 1, 3, 5},
+				giveSortedErr: errors.New("giveSorted error"),
+			},
+		},
+		{
+			name: "Test 3: When sortedStakers is nil",
+			args: args{
+				sortedStakers: nil,
+			},
```
ashish10677, line 444: No check for the time if the error is "gas limit reached"?

ashish10677 (changes requested):

cmd/dispute_test.go
Yashk767: It was not added as It gets stuck in the recursion loop once it enters.

ashish10677: It should not get stuck if the code is correct imo. Can you put a debugger and check why exactly it is getting stuck. If it is getting stuck, we might need to revisit the code.

ashish10677 (changes requested):

cmd/dispute_test.go
Yashk767: Added the test case for recursive call now.

cmd/dispute_test.go
```diff
@@ -405,3 +405,70 @@ func TestHandleDispute(t *testing.T) {
...
+		},
+		{
+			name: "Test 4: When error is gas limit reached",
+			args: args{
+				sortedStakers: []uint32{2, 1, 3, 5},
```
ashish10677, line 449: Can you increase the number of sortedStakers for this test. Anywhere around 15-20 will be good. You can keep this test and add one more test with more number of stakers.

ashish10677 (approved):
LGTM

PR #573, opened 2022-02-14
Added tests for utils/common.go

### LINKED ISSUE #523: Add tests for utils/common.go

shekhar2807:
-Tests added for utils/common.go
-fixes #523


addresses.json
```diff
@@ -9,4 +9,4 @@
   "RAZOR": "0xA792597Ec3C7C02CBf911A7131ceE46B3a002d97",
   "StakedTokenFactory": "0xd10C19DD427ea6D1245E4637b843e84836b100C6",
   "RandomNoManager": "0xb605E2E7134DD3B8BC1C06494bf34cb121ebAc29"
-}
+}
```
SkandaBhat, line 12: unnecessary change. please revert.

cmd/struct-utils.go
```diff
@@ -33,23 +33,23 @@ func (u Utils) ViperWriteConfigAs(path string) error {
 }
 
 func (u Utils) GetEpoch(client *ethclient.Client) (uint32, error) {
-	return utils.GetEpoch(client)
+	return utils.UtilsInterface.GetEpoch(client)
```
SkandaBhat, line 36: Can we not instantiate utils.UtilsInterface and use that on every line? There is too much code duplication. Please fix.

SkandaBhat (changes requested):
please fix the comments

SkandaBhat:
@shekhar2807 there are conflicts in this PR

utils/common.go
```diff
@@ -16,23 +15,23 @@ import (
...
 	return client
 }
 
-func FetchBalance(client *ethclient.Client, accountAddress string) (*big.Int, error) {
+func (*UtilsStruct) FetchBalance(client *ethclient.Client, accountAddress string) (*big.Int, error) {
 	address := common.HexToAddress(accountAddress)
 	coinContract := UtilsInterface.GetTokenManager(client)
 	opts := UtilsInterface.GetOptions()
-	return coinContract.BalanceOf(&opts, address)
+	return UtilsInterface.BalanceOf(coinContract, &opts, address)
```
ashish10677, line 31: `UtilsInterface.BalanceOf` makes less sense. Can you make it `CoinInterface.BalanceOf` or something similar?

utils/common.go
```diff
@@ -47,38 +46,38 @@ func GetDelayedState(client *ethclient.Client, buffer int32) (int64, error) {
...
 	return int(tx.Status)
 }
 
-func WaitForBlockCompletion(client *ethclient.Client, hashToRead string) int {
+func (*UtilsStruct) WaitForBlockCompletion(client *ethclient.Client, hashToRead string) int {
 	timeout := core.BlockCompletionTimeout
 	for start := time.Now(); time.Since(start) < time.Duration(timeout)*time.Second; {
 		log.Debug("Checking if transaction is mined....")
-		transactionStatus := checkTransactionReceipt(client, hashToRead)
+		transactionStatus := UtilsInterface.CheckTransactionReceipt(client, hashToRead)
```
ashish10677, line 62: Can you make `UtilsInterface.CheckTransactionReceipt` as `TransactionUtils.CheckTransactionReceipt`?

utils/common.go
```diff
@@ -97,8 +96,8 @@ func IsFlagPassed(name string) bool {
 	return found
 }
 
-func CheckEthBalanceIsZero(client *ethclient.Client, address string) {
-	ethBalance, err := client.BalanceAt(context.Background(), common.HexToAddress(address), nil)
+func (*UtilsStruct) CheckEthBalanceIsZero(client *ethclient.Client, address string) {
+	ethBalance, err := Client.BalanceAt(client, context.Background(), common.HexToAddress(address), nil)
```
ashish10677, line 100: Why is client being passed as an extra parameter?
My understanding was it'll be `Client.BalanceAt(...)` and the `Client` will contain `client` which is essentially `*ethclient.Client`.

ashish10677 (changes requested):

utils/common.go
shekhar2807: This function is in utils/common.go only , so for changing this it would change whole pattern and this does not come under any native package. I think we can keep it as it is , let me know if you feel otherwise. cc @ashish10677

utils/common.go
ashish10677: Change Client to ClientInterface. So the call becomes, `ClientInterface.BalanceAt(...)`

ashish10677 (approved):

SkandaBhat (approved):

PR #579, opened 2022-02-21
Add Error Message and codes for addStake and claimBounty.

### LINKED ISSUE #40: Implement custom errors
# Rationale

All the error messages (and probably error codes) should be kept in one file. This will help to make the code look better and be more readable. Also, in future, if we decide to change any message, we can do it in one place rather than doing it in many files.

Also, implement an interface for user-defined errors.

# Implementation

An `error.go` file can be created which contains all the error messages and their respective error code.
An interface can be created that implements all the user defined error messages and their respective error code.

ashish10677:
# Description

The error codes and messages are consolidated in the errors.go file now.
The `GetError` function is now used to get the error message and code.

Fixes #40 

## Type of change

- [ ] Code cleanup

### DIFF
--- a/cmd/addStake.go
+++ b/cmd/addStake.go
@@ -31,18 +31,18 @@ func initialiseStake(cmd *cobra.Command, args []string) {
 
 func (*UtilsStruct) ExecuteStake(flagSet *pflag.FlagSet) {
 	config, err := cmdUtils.GetConfigData()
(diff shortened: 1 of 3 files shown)


ashish10677:
@SkandaBhat I'm not very sure this change would add any value right now. On the contrary it is an overhead to add so many error message and codes. Please let me know what do you think.

PR #616, opened 2022-03-14
ETA added for withdraw

### LINKED ISSUE #611: Withdraw lock do not wait for lock
On withdraw, print error message along with ETA to withdraw and exit.

shekhar2807:
# Description

-ETA added for withdraw . Now does not wait for the epoch where staker can withdraw the stake.

Fixes #611 

Screenshot 2022-03-14 at 6 45 20 PM


SkandaBhat (changes requested):
We also need to give approx time in seconds. "Please wait for 1 epoch! (approximately X seconds)"

shekhar2807:
> We also need to give approx time in seconds. "Please wait for 1 epoch! (approximately X seconds)"

@SkandaBhat 
Screenshot 2022-03-16 at 2 02 47 PM

cmd/withdraw.go
```diff
@@ -99,21 +99,22 @@ func (*UtilsStruct) WithdrawFunds(client *ethclient.Client, account types.Accoun
...
-		epoch, err = razorUtils.GetUpdatedEpoch(client)
+	waitFor := big.NewInt(0).Sub(lock.WithdrawAfter, big.NewInt(int64(epoch)))
+	if waitFor.Cmp(big.NewInt(0)) > 0 {
+		timeRemaining := (time.Duration(int64(uint32(waitFor.Int64()))*core.EpochLength*razorUtils.CalculateBlockTime(client)) * time.Second) / (1000000000)
+		value, err := strconv.Atoi(big.NewInt(int64(timeRemaining)).String())
 		if err != nil {
-			log.Error("Error in fetching epoch")
 			return core.NilHash, err
 		}
+		log.Info("Withdrawal period not reached. Cannot withdraw now, please wait for", waitFor, "epochs!", "(approximately", razorUtils.SecondsToHuman(value), ")")
```
ashish10677, line 109: Can you just add a small check to see if the waitFor value is 1, then print epoch instead of epochs.
Also, use `log.Infof` here instead of `log.Info`

utils/time.go
```diff
@@ -0,0 +1,48 @@
...
+func plural(count int, singular string) (result string) {
+	if (count == 1) || (count == 0) {
+		result = strconv.Itoa(count) + " " + singular + " "
+	} else {
+		result = strconv.Itoa(count) + " " + singular + "s "
+	}
+	return
+}
+
+func (*UtilsStruct) SecondsToHuman(input int) (result string) {
```
ashish10677, line 17: Change this name to `SecondsToReadableTime` or `SecondsToTime`.

utils/time.go
```diff
@@ -0,0 +1,48 @@
+package utils
+
+import (
+	"math"
+	"strconv"
+)
+
+func plural(count int, singular string) (result string) {
```
ashish10677, line 8: remove `result` as return variable here and add return result at last.
This is not wrong, but we are not using this format anywhere else imo. Just to maintain consistency.

ashish10677 (changes requested):

cmd/withdraw.go
shekhar2807: Done @ashish10677

utils/time.go
shekhar2807: Done @ashish10677

utils/time.go
shekhar2807: Done @ashish10677

ashish10677 (approved):
LGTM

PR #628, opened 2022-03-22
Logging mechanism improvement

### LINKED ISSUE #581: Logging mechanism to be reviewed and overhauled
# Rationale

Currently the logging implementation is writing to a file `razor.log`. This is a cause for concern as a staker might run multiple stakers on the same machine, and the same file is being used by all the processes. There is a chance it might cause a failure

# Implementations (needs to be analysed)

1. To be explicit in requiring logfile before logging; otherwise log to stdout by default
~~2. Implement a system where we ensure that when a log file is being used by a process, no other process can use the log file.~~
~~3. Explore other patterns.~~
~~4. Verify if logging library addresses this issue (logrus)~~

shekhar2807:
# Description

- New flag has been added in every command named as logFile where user can pass the file name where they want to store       the logs of that command . If the flag is not passed output will be only printed on terminal and will not be stored anywhere.

Fixes #581


logger/logger.go
```diff
@@ -53,6 +39,34 @@ func init() {
...
+			MaxAge:     30,
+		}
+
+		out := os.Stderr
+		mw := io.MultiWriter(out, lumberJackLogger)
+		standardLogger.Formatter = &logrus.JSONFormatter{}
+		standardLogger.SetOutput(mw)
+
+	} else {
+		_, err := path.PathUtilsInterface.GetLogFilePath(fileName)
```
ashish10677, line 62: If log fileName isn't provided just write to stdout, you don't need to fetch the path.

cmd/addStake.go
```diff
@@ -120,6 +129,7 @@ func init() {
 		VoteAutomatically bool
 		Rogue             bool
 		RogueMode         []string
+		LogFile           string
```
ashish10677, line 132: Instead of explicitly adding a logFile flag in every command, can you not add it in global way so that all commands by default have it? Like logLevel, or gasPrice, etc.

ashish10677 (changes requested):

cmd/addStake.go
```diff
@@ -30,6 +30,7 @@ func initialiseStake(cmd *cobra.Command, args []string) {
 }
 
 func (*UtilsStruct) ExecuteStake(flagSet *pflag.FlagSet) {
+	cmdUtils.AssignLogFile(flagSet)
```
ashish10677, line 33: `logger.Address = address` can be present in AssignLogFile function itself I think

README.md
```diff
@@ -764,6 +764,21 @@ In the above example for the collection `ethCollectionMean`, new custom job havi
...
+$ ./razor addStake --address <address> --value <value> --logFile stakingLogs
+```
+_The logs for above command will be stored in stakingLogs.log file in .razor directory_
+
+```
+$ ./razor delegate --address <address> --value <value> --pow <power> --stakerId <staker_id> --logFile delegationLogs
+```
+_The logs for above command will be stored in delegationLogs.log file in .razor directory_
+
+
```
ashish10677, line 781: Add a note here regarding multiple commands running with same log file

ashish10677:
@SkandaBhat according to this PR, the logFile will be always stored in `.razor` directory. I don't think that is the requirement. The flag `--logFile` must take the entire path and create a file at that path. Not specifically in the `.razor` directory.

README.md
```diff
@@ -764,6 +764,22 @@ In the above example for the collection `ethCollectionMean`, new custom job havi
...
 ```
 
+### Logs
+
+User can pass a seperate flag --logFile followed with any name for log file along with command. The logs will be stored in ```.razor``` directory.
+
+```
+$ ./razor addStake --address <address> --value <value> --logFile stakingLogs
+```
+_The logs for above command will be stored in stakingLogs.log file in .razor directory_
```
ashish10677, line 774: Mention the full path here

ashish10677 (changes requested):

README.md
shekhar2807: Done @ashish10677

ashish10677 (approved):

PR #694, opened 2022-05-04
Updated install script

Shrikant1212:
# Description

Updated Install script for binary install

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
-version="${VERSION:-v1.0.0}"
+set -e
+version="${VERSION:-v0.0.0}"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
-version="${VERSION:-v1.0.0}"
+set -e
```
ashish10677, line 2: From my understanding, you are trying to pick up version from the environment variable right?
That's what the PR does?

Shrikant1212: Yeah, I'm reading Version from Env Variable, then that will be used in downloading binary.

ashish10677 (approved):
LGTM

PR #702, opened 2022-05-12
Update install script with stabe release as default version

Shrikant1212:
# Description

Updated install script to use latest stable version as default if no Version Environment is passed.


ashish10677 (approved):
lgtm

PR #719, opened 2022-05-20
updated docker commands in readme

Shrikant1212:
# Description
Updated  docker command to use exec instead of running run commands in readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,8 +120,7 @@ This will cause this particular vote command to run with a gas price of 10.
 
 
 
-
-
+## Razor commands
 
 ### Create Account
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,65 +28,34 @@ Check installation
 razor -v
 ```
 >**_NOTE:_** To install the version you want, you can set VERSION:<git-tag> environment variable before running above command.
-### Docker quick start
+##Docker quick start
```
ashish10677, line 31: Put a space between the 2nd hash and `Docker`

ashish10677 (changes requested):

PR #720, opened 2022-05-20
Dispute Fix

### LINKED ISSUE #718: Dispute Fails
@Yashk767 Please add more details here.

Yashk767:
# Description
Dispute was failing as same votes were getting appended to sortedVotes array for a collection which resulted in wrong calculation of total influence revealed. Fixed that by re-correcting the check of contains to `big.int(vote)` is present in `*big.int[]votes` array.

Fixes #718


ashish10677 (approved):
LGTM

PR #743, opened 2022-05-30
Separate directories to store data and keystore files in .razor

### LINKED ISSUE #740: Store Keystore and data files in a separate directory in .razor
Currently we have keystore and data files associated to an address in `.razor` directory.
When we try to fetch private key from keystore file, razor go iterates over files in `.razor` using `accounts.Accounts()` and search for the corresponding address in that file,

This is unsafe as, if any file having `{"address": }` in the first line and the value to that address is an valid address, `accounts.Accounts` will pick it up and we might get wrong data.

Yashk767:
# Description

Now data files(commit, propose and dispute data) and keystore files are stored in a separate directories in `.razor`. 

Fixes #740

### DIFF
--- a/path/path.go
+++ b/path/path.go
@@ -62,7 +62,7 @@ func (PathUtils) GetCommitDataFileName(address string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	dataFileDir := pathPkg.Join(razorDir, "dataFiles")
+	dataFileDir := pathPkg.Join(razorDir, "data_files")
 	if _, err := OSUtilsInterface.Stat(dataFileDir); OSUtilsInterface.IsNotExist(err) {
 		mkdirErr := OSUtilsInterface.Mkdir(dataFileDir, 0700)
 		if mkdirErr != nil {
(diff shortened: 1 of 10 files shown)


path/path.go
```diff
@@ -59,7 +59,15 @@ func (PathUtils) GetCommitDataFileName(address string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return homeDir + "/" + address + "_CommitData.json", nil
+	dataFileDir := homeDir + "/dataFiles"
```
ashish10677, line 62: Can you use path.Join for joining paths everywhere?

ashish10677 (changes requested):

path/path.go
```diff
@@ -36,68 +39,68 @@ func (PathUtils) GetLogFilePath(fileName string) (string, error) {
...
 
 //This function returns the file name of commit data file
 func (PathUtils) GetCommitDataFileName(address string) (string, error) {
-	homeDir, err := PathUtilsInterface.GetDefaultPath()
+	razorDir, err := PathUtilsInterface.GetDefaultPath()
 	if err != nil {
 		return "", err
 	}
-	dataFileDir := homeDir + "/dataFiles"
+	dataFileDir := pathPkg.Join(razorDir, "dataFiles")
```
ashish10677, line 65: Keep the directory names as underscore separated, not in camel case.
eg: `data_files`

ashish10677 (changes requested):

path/path.go
ashish10677: Yes

ashish10677 (approved):

PR #761, opened 2022-06-17
IOUTIL package deprecated

### LINKED ISSUE #757: ioutil readall function is deprecated after go 1.16

shekhar2807:
# Description

- IOUTIL package has been deprecated after go 1.16
- This has been removed in this PR and affected functions has been changed to the latest supporting packages.

Fixes #757


utils/struct-utils.go
```diff
@@ -394,21 +403,6 @@ func (a ABIStruct) Pack(parsedData abi.ABI, name string, args ...interface{}) ([
...
-	return ioutil.ReadAll(body)
-}
-
-//This function is used to read the file
-func (i IoutilStruct) ReadFile(filename string) ([]byte, error) {
-	return ioutil.ReadFile(filename)
-}
-
-//This function is used to write data to file
-func (i IoutilStruct) WriteFile(filename string, data []byte, perm fs.FileMode) error {
```
ashish10677, line 408: You've removed WriteFile, is it not added?

ashish10677 (changes requested):

utils/struct-utils.go
shekhar2807: It's already there is struct utils. @ashish10677 
https://github.com/razor-network/razor-go/blob/41d326ad9374af7042931ddedf671ad20fbb1039/utils/struct-utils.go#L338

Earlier we had writeFile from both the packages inutile and os .

ashish10677: Ok cool.

ashish10677 (approved):
LGTM

PR #763, opened 2022-06-22
crypto/rand is used instead of math/rand

### LINKED ISSUE #753: Use crypto/rand instead of math/rand

shekhar2807:
# Description

- crypto/rand is used instead of math/rand in cmd/propose.go and utils/math.go
- math/rand is used in cmd/dispute.go as it uses rand.Perm function which is provided by only math/rand package.

Fixes #753

### DIFF
--- a/cmd/propose.go
+++ b/cmd/propose.go
@@ -2,6 +2,7 @@
 package cmd
 
 import (
+	"crypto/rand"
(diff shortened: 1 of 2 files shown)


ashish10677 (approved):
LGTM

PR #788, opened 2022-06-30
--Pow flag is removed

### LINKED ISSUE #772: Power Flag improvements
- [ ] Create a flag which accepts value in wei

--weiRazor true — value <razor value in wei>

shekhar2807:
# Description

- --pow flag is removed.
- --weiRazor is added which accepts bool, if this is passed user can enter value in Wei.

```--weiRazor true — value <razor value in wei>```

Fixes #772

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,28 +194,22 @@ Example:
 $ ./razor addStake --address 0x5a0b54d5dc17e0aadc383d2db43b0a0d3e029c4c --value 1000
 ```
 
-_Note: --pow flag is used to stake floating number stake_
+_Note: --weiRazor flag can be passed to provide values in wei_
 
-_Note: Formula for calculating pow: (value \* (10**18)) / (10**x) where x is no of decimal places and value is integer_
-
-_The value of pow is : 18 - x here_
-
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -292,19 +286,19 @@ If you want to become a delegator use the `delegate` command. The staker whose `
...
 ```
-docker exec -it razor-go razor delegate --address <address> --value <value> --pow <power> --stakerId <staker_id>
+docker exec -it razor-go razor delegate --address <address> --value <value> --weiRazor <bool> --stakerId <staker_id>
 ```
 
 Example:
 
 ```
-$ ./razor delegate --address 0x5a0b54d5dc17e0aadc383d2db43b0a0d3e029c4c --value 1000 --pow 10 --stakerId 1
+$ ./razor delegate --address 0x5a0b54d5dc17e0aadc383d2db43b0a0d3e029c4c --value 1000 --weiRazor --false --stakerId 1
```
ashish10677, line 301: This should be `false` not `--false`

ashish10677 (changes requested):

README.md
shekhar2807: Done @ashish10677

PR #818, opened 2022-07-13
added minStake check in Initiate commit, reveal and propose

### LINKED ISSUE #816: Check for minstake and return on commit, reveal, propose
- [ ] Check for minstake and return on commit, reveal, propose
- [ ] Remove mistake check from outside the switch case

shekhar2807:
# Description

- Added minStake check in InitiateCommit, InitiateReveal and InitiatePropose
- Removed minStake check from HandleBlock
- Fixed Tests according to the above changes

Fixes #816


ashish10677 (approved):
lgtm

PR #823, opened 2022-07-15
randomSortedProposedBlockIds fix

### LINKED ISSUE #822: `randomSortedProposedBlockIds` should be randomizing the sortedProposedBlockIds array
#### Describe the bug
```
randomSortedProposedBlockIds := rand.Perm(len(sortedProposedBlockIds)) //returns random permutation of integers from 0 to n-1
```

This line returns the values from 0-n-1 as an array
But rather we need the permutation of sortedProposedBlockIds array

#### Expected behaviour
Basically if sortedProposedBlockIds = [45,65,23,64,12]
then ideally, randomSortedProposedBlockIds = [23,64,12,65,45]

#### Actual behaviour
But the current code gives us [3, 2, 4, 1, 0]

shekhar2807:
# Description

- Random sortedProposedBlocKIds fix
- Test added for shuffle function 

Fixes #822

### DIFF
--- a/cmd/dispute_test.go
+++ b/cmd/dispute_test.go
@@ -1150,6 +1150,7 @@ func BenchmarkHandleDispute(b *testing.B) {
 	var blockNumber *big.Int
 	var rogueData types.Rogue
 	var blockManager *bindings.BlockManager
+	var randomSortedPorposedBlockIds []uint32
 
 	table := []struct {
 		numOfSortedBlocks uint32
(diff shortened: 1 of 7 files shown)


cmd/dispute_test.go
```diff
@@ -275,9 +282,10 @@ func TestHandleDispute(t *testing.T) {
 		{
 			name: "Test 8: When DisputeBiggestStakeProposed transaction fails",
 			args: args{
-				sortedProposedBlockIds: []uint32{3, 1, 2, 0, 4},
-				biggestStake:           big.NewInt(1).Mul(big.NewInt(5356), big.NewInt(1e18)),
-				biggestStakeId:         2,
+				sortedProposedBlockIds:       []uint32{3, 1, 2, 0, 4},
+				randomSortedProposedBlockIds: []uint32{2, 3, 1, 4, 0},
```
ashish10677, line 286: Can you use some other number than 0-5?

utils/math_test.go
```diff
@@ -890,3 +890,57 @@ func TestGetRogueRandomMedianValue(t *testing.T) {
...
+func TestShuffle(t *testing.T) {
+	type args struct {
+		slice []uint32
+	}
+	tests := []struct {
+		name string
+		args args
+		want bool
+	}{
+		{
```
ashish10677, line 903: Add more tests, like when array is blank, etc.

ashish10677 (changes requested):

cmd/dispute_test.go
shekhar2807: Done @ashish10677

ashish10677 (approved):

PR #847, opened 2022-07-19
updated readme for running razor commands in background

### LINKED ISSUE #836: Update README for running razor-go in background
Now that we have removed the feature to accept password as a file, the README must be updated so that users can run their stakers in background.

Shrikant1212:
Fixes #836

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,20 +344,23 @@ docker exec -it razor-go razor vote --address <address>
 ```
 
 To run vote command in background we can use `tmux`, a linux utility for that
-```
-# Create session 
-tmux new -s razor-go
 
-# Run vote command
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -338,9 +343,20 @@ docker
 docker exec -it razor-go razor vote --address <address>
 ```
 
-run vote command in background
+To run vote command in background we can use `tmux`, a linux utility for that
 ```
-docker exec -it -d razor-go razor vote --address <address> 
+# Create session
```
ashish10677, line 348: Can you put these as bullet or numeric points?
image
This doesn't look very good.

ashish10677 (changes requested):

ashish10677 (approved):

PR #873, opened 2022-07-20
Added valid ERC20 check for addresses

### LINKED ISSUE #837: Check for address validity everywhere using a validation helper
Ethereum addresses can be validated when user enters their address in any command. If the address doesn't comply with ethereum (like) address, we don't need to proceed and razor-go should throw an error.

shyam-patel-kira:
fixes #837


utils/common.go
```diff
@@ -92,6 +92,12 @@ func CheckError(msg string, err error) {
 	}
 }
 
+func IsValidERC20Address(address string) {
```
ashish10677, line 95: @shyam-patel-kira please add tests for this function.
And it would be better if this function did not fatal here, rather it returned an error.
That error can be handled wherever this function is called.

ashish10677 (changes requested):

utils/common.go
shyam-patel-kira: okay

shyam-patel-kira: Done

ashish10677 (approved):

PR #874, opened 2022-07-21
Lumberjack constants

### LINKED ISSUE #832: Move lumberjack (logger.go) constants to config
https://github.com/razor-network/razor-go/blob/982502c3c715b86483809aa99d303f4afc7144ad/logger/logger.go#L49

Move these constant values to config, so that the user can decide the log rotation cycle.

Yashk767:
# Description

Moved lumberjack constants to config.

Fixes #832


cmd/cmd-utils.go
```diff
@@ -23,7 +23,7 @@ func (*UtilsStruct) GetEpochAndState(client *ethclient.Client) (uint32, int64, e
 	if err != nil {
 		return 0, 0, err
 	}
-	bufferPercent, err := stringUtils.ParseInt(bufferPercentString)
+	bufferPercent, err := stringUtils.ParseInt64(bufferPercentString)
```
ashish10677, line 26: Any specific reason to change ParseInt to ParseInt64?

Yashk767: Before the change, `ParseInt(string s)` was parsing string s to int64 but still had name ParseInt not ParseInt64. 

In this PR, the `lumberjack constants` are of type int, so there is a requirement of parsing string to int. For which I needed a function ParseInt() which parses string to int.

ashish10677: Ok

PR #898, opened 2022-07-29
Fix: Move old features to the new main

### LINKED ISSUE #860: move old features to new main branch

shekhar2807:
# Description

- GetIteration concurrency
- Moved chainId to config
- Added compress parameter in logger.go
- Godoc documentation for utils, logger, core, metrics

Fixes #860

### DIFF
--- a/cmd/config-utils.go
+++ b/cmd/config-utils.go
@@ -20,6 +20,7 @@ func (*UtilsStruct) GetConfigData() (types.Configurations, error) {
 		LogFileMaxSize:     0,
 		LogFileMaxBackups:  0,
 		LogFileMaxAge:      0,
+		Compress:           "",
 	}
 	provider, err := cmdUtils.GetConfig("provider")
 	if err != nil {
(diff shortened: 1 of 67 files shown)


Yashk767 (dismissed):
LGTM

ashish10677 (changes requested):

logger/logger.go
```diff
@@ -65,6 +67,7 @@ func InitializeLogger(fileName string) {
 			MaxSize:    logFileMaxSize,
 			MaxBackups: logFileMaxBackups,
 			MaxAge:     logFileMaxAge,
+			Compress:   true,
```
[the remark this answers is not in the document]
shekhar2807: Done @ashish10677

cmd/config-utils.go
```diff
@@ -20,6 +20,7 @@ func (*UtilsStruct) GetConfigData() (types.Configurations, error) {
 		LogFileMaxSize:     0,
 		LogFileMaxBackups:  0,
 		LogFileMaxAge:      0,
+		Compress:           "",
```
ashish10677, line 23: Is there any problem keeping this as a bool instead of a string?

shekhar2807: Yes, I tried using that also . It was throwing some errors . As far as String is concerned I checked for every possibility. It works for every case.

shekhar2807: Let me try that again, will update it here .

shekhar2807: if compress != "" {
 		viper.Set("compress", compress)
} . This if statement cause the issue if compress is stated as boolean . String solves this issue and make it easier to implement. @ashish10677

ashish10677 (approved):
Lgtm

PR #909, opened 2022-08-31
CI for private Releases

### LINKED ISSUE #765: Figure out CI for Private Releases without contract release
# Rationale

# Implementation

shyam-patel-kira:
fixes #765

### DIFF
--- a/export.sh
+++ b/export.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+
+set -e -o pipefail
+
+mkdir -p ./pkg/bindings
+
+generate_binding() {
+  contract=$(echo $1 | awk '{print $1}')
+  contractType=$(echo $1 | awk '{print $2}')
+  go_source=$(echo $1 | awk '{print $3}')
+  echo "Generating binding for (${contract})"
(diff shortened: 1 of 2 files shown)


export.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
ashish10677, line 1: Instead of adding a new file, can you modify generate-bindings.sh?

export.sh
```diff
@@ -0,0 +1,50 @@
...
+  "StakedToken stakedToken.go"
+  "RAZOR RAZOR.go"
+  "AccessControl accessControl.go"
+  "ERC20 erc20.go"
+  "ERC165 erc165.go"
+)
+
+commitId=$1
+rm -rf contract-abi
+git clone https://github.com/razor-network/contracts.git
```
ashish10677, line 34: This will clone the contracts in razor-go itself right?

ashish10677 (changes requested):

export.sh
shyam-patel-kira: Yes it will clone contracts export abis for the given commitId and remove it

export.sh
shyam-patel-kira: okay, can modify it for public and private releases

PR #913, opened 2022-09-01
Hotfix-proposed data

### LINKED ISSUE #910: While running 2 nodes, give sorted is getting called.
#### System information

- **OS & Version**: Ubuntu
- **Go version:** go 1.17
- **razor-go version:** mainnet-stable

#### Describe the bug
When running 2 nodes, give sorted is getting called, because razor-go tries to fetch the last epoch's value and not this epoch's value.

1. Run 2 nodes
2. Check if values of last proposed epoch are correct.

Yashk767:
# Description

- Updated proposed data global variables correctly.
- We save the propose data to proposedDataFile only on successful propose.

Fixes #910

### DIFF
--- a/cmd/dispute.go
+++ b/cmd/dispute.go
@@ -43,10 +43,13 @@ func (*UtilsStruct) HandleDispute(client *ethclient.Client, config types.Configu
 	}
 	log.Debug("Biggest Stake: ", biggestStake)
 
-	medians, revealedCollectionIds, revealedDataMaps, err := cmdUtils.GetLocalMediansData(client, account, epoch, blockNumber, rogueData)
+	locallyCalculatedData, err := cmdUtils.GetLocalMediansData(client, account, epoch, blockNumber, rogueData)
 	if err != nil {
 		return err
 	}
+	medians := locallyCalculatedData.MediansData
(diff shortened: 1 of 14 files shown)


cmd/dispute.go
```diff
@@ -224,14 +221,17 @@ CalculateMedian:
...
+			RevealedDataMaps:      revealedDataMaps,
+			Epoch:                 epoch,
+		})
 	}
 
 	log.Debug("Locally calculated data:")
-	log.Debugf("Medians: %d", _mediansData)
-	return _mediansData, _revealedCollectionIds, _revealedDataMaps, nil
+	log.Debugf("Medians: %d", globalProposedDataStruct.MediansData)
+	return globalProposedDataStruct.MediansData, globalProposedDataStruct.RevealedCollectionIds, globalProposedDataStruct.RevealedDataMaps, nil
```
ashish10677, line 234: You can return globalProposedDataStruct directly right?

ashish10677 (changes requested):

cmd/dispute.go
Yashk767: globalProposedDataStruct has extra field `epoch`.
Should we return `epoch` as well from `GetLocalMediansData` ?

ashish10677 (approved):

PR #917, opened 2022-09-02
Fetched Last proposed from contracts

### LINKED ISSUE #912: Get last proposed epoch should be called from the smart contract
Right now, we use logbloom to get last proposed epoch. We need to fetch it from public mapping from the smart contract instead, since logbloom could have accuracy problems.

https://github.com/razor-network/contracts/blob/master/contracts/Core/storage/BlockStorage.sol#L13

shekhar2807:
# Description

- Fetched getEpochLastProposed from contracts
- Modified and added tests fot it.

Fixes #912


ashish10677 (approved):
lgtm

Yashk767 (approved):

PR #921, opened 2022-09-05
Hotfix-giveSorted

### LINKED ISSUE #911: giveSorted fixes
#### Describe the bug
Once givesorted tx is confirmed, the next givesorted should know where it has to start from. 

#### Expected behaviour

- [x] if there is no need of givesorted it should not call it.

- [x] if prev giveSorted was incorrect, then reset dispute and start again.

- [ ] if prev giveSorted was correct, do not call giveSorted again. even if finalize dispute fails.

https://razorscan.io/tx/0xb962064a6f3d3a76852dc0e9c94f1943e015908e0e76628e4ea4053dde551575

https://razorscan.io/tx/0x2cade48c0cc69e7180b17bf870a68fee827243e4937933c534bc6440578c5f05

- [x] if one of the givesorted fails, then reset and try again (in recursion)
- [x] If govesorted fails, do not call finalizeDispute

Run 2 nodes, the other node tries to give Sorted.

- [x] Scenario 1: giveSorted goes thru in 1 go
- [x] Scenario 2: giveSorted exceeds gas limit, is split into multiple calls and executed sequentially
- [x] Scenario 3: giveSorted exceeds gas limit, but one of the calls fails.
//	eg: chunks of 10, among 100. Here if first chunk fails, then second onwards should not be called. This does not happen.

- [x] Scenario 4: redundant nodes run giveSorted. 
// a. giveSorted chunk is called by first node a and succeeds.
// b. node b is not aware of this, and calls giveSorted for the same chunk, and the transaction fails.
// c. node b thinks it is an invalid giveSorted, and calls resetDispute.
// d. node a is not able to finalise dispute because it was reset by node b. 

- [x] Scenario 5: node is restarted in dispute state.
// Currenlty, giveSorted is called again for all. If giveSorted is called previously, then resetDispute is called.

- [x]  Scenario 6: node a does giveSorted but tx is pending
// if tx is not confirmed, node b also does giveSorted
// node b's tx fails. node a's tx is successful.
// Since we are validating giveSorted from the blockchain, node b will not call ResetDispute.

- [x] Scenario 7: onchain state is invalid and both nodes call resetDispute at the same time.
// node a: resetDispute (n: 1), giveSorted (n: 2)
// node b: resetDispute (n: 2), giveSorted (n: 3)

- [ ]  Scenario 8: node a givesSorted for block 1, node b does the same for block 7
// node should check on chain state and should not continue dispute for block 7 when block on is ongoing. It should continue dispute for block 1.
// node b will end up never disputing for block 7. we should recheck for disputes multiple times in the state.

// Solution:
- [x]  1. calculate nonce from confirmed transactions and not pending transactions
- [x]  2. Validate disputes status before every dispute-related call.-

The index of the smallest sorted values should be greater to your LVV.

- [] a. node should be aware of dispute status. If the block is already disputed, then we should not go thru with dispute.
- [x]  b. node should be aware of ongoing dispute, and continue it.
- [x]  3. Repeat dispute calculation off chain multiple times throughout dispute state.

fetch from disputes struct: https://github.com/razor-network/oracle-contracts/blob/master/contracts/Core/storage/BlockStorage.sol#L7

Yashk767:
# Description

- Reset Dispute is only after `FinalizeDispute`, removed from other dispute types.
- `ResetDispute` called irrespective of `FinalizeDispute` Txn status.

Fixes #911

### DIFF
--- a/cmd/dispute.go
+++ b/cmd/dispute.go
@@ -6,7 +6,7 @@ import (
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
-	types2 "github.com/ethereum/go-ethereum/core/types"
+	Types "github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/ethclient"
 	solsha3 "github.com/miguelmota/go-solidity-sha3"
 	"math/big"
@@ -21,14 +21,12 @@ import (
 
 var (
 	giveSortedLeafIds []int
-	disputedFlag      bool
 )
 
 //blockId is id of the block
 
 //This function handles the dispute and if there is any error it returns the error
 func (*UtilsStruct) HandleDispute(client *ethclient.Client, config types.Configurations, account types.Account, epoch uint32, blockNumber *big.Int, rogueData types.Rogue) error {
-	disputedFlag = false
 
 	sortedProposedBlockIds, err := razorUtils.GetSortedProposedBlockIds(client, epoch)
 	if err != nil {
@@ -66,7 +64,8 @@ func (*UtilsStruct) HandleDispute(client *ethclient.Client, config types.Configu
 			log.Error(err)
 			continue
 		}
-		log.Debug("Proposed block ", blockId, proposedBlock)
+		log.Debug("Block ID: ", blockId)
+		log.Debug("Proposed block ", proposedBlock)
 
 		//blockIndex is index of blockId in sortedProposedBlock
 		blockIndex := utils.IndexOf(sortedProposedBlockIds, uint32(blockId))
@@ -97,7 +96,6 @@ func (*UtilsStruct) HandleDispute(client *ethclient.Client, config types.Configu
 
 			//If dispute happens, then storing the bountyId into disputeData file
 			if WaitForBlockCompletionErr == nil {
-				disputedFlag = true
 				err = cmdUtils.StoreBountyId(client, account)
 				if err != nil {
 					log.Error(err)
@@ -121,7 +119,6 @@ func (*UtilsStruct) HandleDispute(client *ethclient.Client, config types.Configu
 
 			//If dispute happens, then storing the bountyId into disputeData file
 			if WaitForBlockCompletionErr == nil {
-				disputedFlag = true
 				err = cmdUtils.StoreBountyId(client, account)
 				if err != nil {
 					log.Error(err)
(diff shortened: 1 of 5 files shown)


cmd/dispute.go
```diff
@@ -304,59 +289,83 @@ func (*UtilsStruct) CheckDisputeForIds(client *ethclient.Client, transactionOpts
...
+			err = cmdUtils.StoreBountyId(client, account)
+			if err != nil {
+				return err
+			}
+		} else {
+			log.Error("Error in WaitForBlockCompletion for FinalizeDispute: ", WaitForBlockCompletionErr)
 		}
 	}
+
+	//Resetting dispute irrespective of FinalizeDispute transaction status
```
ashish10677, line 337: Are you sure we have to reset dispute irrespective of whether FinalizeDispute goes through or not?

cmd/dispute.go
```diff
@@ -304,59 +289,83 @@ func (*UtilsStruct) CheckDisputeForIds(client *ethclient.Client, transactionOpts
...
 	if err != nil {
+		log.Error("Error in calling GiveSorted: ", err)
 		if err.Error() == errors.New("gas limit reached").Error() {
-			log.Error("Error in calling GiveSorted: ", err)
 			mid := len(sortedValues) / 2
-			GiveSorted(client, blockManager, txnOpts, epoch, leafId, sortedValues[:mid])
-			GiveSorted(client, blockManager, txnOpts, epoch, leafId, sortedValues[mid:])
+			err1 := GiveSorted(client, blockManager, txnOpts, epoch, leafId, sortedValues[:mid])
+			err2 := GiveSorted(client, blockManager, txnOpts, epoch, leafId, sortedValues[mid:])
+			if (err1 == nil && err2 != nil) || (err1 != nil && err2 == nil) {
```
ashish10677, line 364: Shouldn't this be `if err1 != nil || err2 != nil`?
Even if we get one error, that means that we have to reset dispute right?

ashish10677 (changes requested):

cmd/dispute.go
Yashk767: Hmm yes, discussed with @SamAg19. 

After giveSorted, the state of dispute changes so irrespective of FinalizeDispute txn status, we need to reset the state for the next dispute. So that next giveSorted for other proposed block goes through.

cmd/dispute.go
Yashk767: But if we get error in both i.e. `err1 != nil` &&` err2 != nil `, we dont need to reset dispute.

ashish10677 (approved):

shekhar2807 (approved):

PR #926, opened 2022-09-05
Call claimStakerReward only if there reward to claim

### LINKED ISSUE #908: Check reward before calling the `claimStakerReward` function in `claimCommission` command
#### System information

- **OS & Version**: Linux
- **Go version:** go 1.17
- **razor-go version:** v1-mainnet-stable

#### Describe the bug
`claimCommission` command can be called by anyone, and it calls the `claimStakerReward` function of the smart contract without checking if the staker actually has accommodated any commission. 

#### Expected behaviour
The smart contract call must not be made if there is no commission to claim.

#### Actual behaviour
The smart contract call is made regardless of whether the commission is there to claim or not.

#### Backtrace
er

shyam-patel-kira:
fixes #908

### DIFF
--- a/cmd/claimCommission.go
+++ b/cmd/claimCommission.go
@@ -3,6 +3,7 @@ package cmd
 
 import (
 	"github.com/spf13/pflag"
+	"math/big"
 	"razor/core"
 	"razor/core/types"
 	"razor/logger"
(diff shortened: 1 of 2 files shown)


cmd/claimCommission.go
```diff
@@ -38,28 +39,37 @@ func (*UtilsStruct) ClaimCommission(flagSet *pflag.FlagSet) {
...
-	if err != nil {
-		log.Fatal("Error in claiming stake reward: ", err)
+	stakerId, err := razorUtils.GetStakerId(client, address)
+	utils.CheckError("Error in getting stakerId: ", err)
+	callOpts := razorUtils.GetOptions()
+
+	stakerInfo, err := stakeManagerUtils.StakerInfo(client, &callOpts, stakerId)
+	utils.CheckError("Error in getting stakerInfo: ", err)
+
+	if stakerInfo.StakerReward.Cmp(big.NewInt(0)) == 1 {
```
ashish10677, line 49: Instead of `== 1` can you put it as `> 1`. Makes it easier to understand and is consistent with rest of the codebase

ashish10677 (changes requested):

cmd/claimCommission.go
shyam-patel-kira: > 1 would not be possible in any case
should I put it as > 0 ?

shyam-patel-kira: done

ashish10677 (approved):

Yashk767 (approved):

PR #945, opened 2022-09-15
Hotfix-giveSorted fixes

### LINKED ISSUE #911: giveSorted fixes

Yashk767:
# Description

- Nonce calculated from conformed transaction, changed `PendingNonceAt` to `NonceAt`
- Checked Disputes state from blockchain before every dispute action
- Added checks whether block is already disputed or not
- Dispute calculation is on throughout dispute state.

Fixes #911

### DIFF
--- a/cmd/dispute.go
+++ b/cmd/dispute.go
@@ -315,12 +315,12 @@ func (*UtilsStruct) Dispute(client *ethclient.Client, config types.Configuration
 	}
 
 	blockId := sortedProposedBlocks[blockIndex]
-	proposedBlock, err := razorUtils.GetProposedBlock(client, epoch, blockId)
+	updatedProposedBlock, err := razorUtils.GetProposedBlock(client, epoch, blockId)
 	if err != nil {
 		log.Error("Error in getting updated proposed block: ", err)
 		return err
 	}
-	if !proposedBlock.Valid {
+	if !updatedProposedBlock.Valid {
 		log.Debug("Block already disputed, not checking medians dispute again")
 		return errors.New("block already disputed")
 	}
--- a/cmd/dispute.go
+++ b/cmd/dispute.go
@@ -64,7 +64,7 @@ func (*UtilsStruct) HandleDispute(client *ethclient.Client, config types.Configu
 	log.Debug("Shuffled sorted proposed blocks: ", randomSortedProposedBlockIds)
 
 	for _, blockId := range randomSortedProposedBlockIds {
-		proposedBlock, err := razorUtils.GetProposedBlock(client, epoch, uint32(blockId))
+		proposedBlock, err := razorUtils.GetProposedBlock(client, epoch, blockId)
 		if err != nil {
 			log.Error(err)
 			continue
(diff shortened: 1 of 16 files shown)


SkandaBhat:
Closing this as implementation has changed. @Yashk767 please raise a new PR with new implementation.

PR #947, opened 2022-09-20
hotfix-CommitData

### LINKED ISSUE #944: _commitData should not be stored in global memory
If a node commits but does not reveal for a few epochs, old commit data is stored in global memory. When the node reveals, it reveals with old commit data, thereby causing the reveal to fail.

NOTE: This is an edge case.

https://github.com/razor-network/oracle-node/blob/main/cmd/vote.go#L448

Yashk767:
# Description

- Saved commit data to commit file only after successful commit.
- Added epoch check for global commit data during reveal.

Fixes #944

### DIFF
--- a/cmd/vote.go
+++ b/cmd/vote.go
@@ -130,10 +130,10 @@ func (*UtilsStruct) Vote(ctx context.Context, config types.Configurations, clien
 }
 
 var (
-	_commitData      types.CommitData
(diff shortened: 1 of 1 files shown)


ashish10677 (approved):
lgtm

shekhar2807 (approved):

PR #948, opened 2022-09-26
Changed pending nonce at to nonce at.

ashish10677:
# Description

`PendingNonceAt` is changed to `NonceAt`

Fixes [#114](https://github.com/razor-network/bridge-node/issues/114)

## Type of change

Please delete options that are not relevant.

# How Has This Been Tested?

- [x] Tested on local hardhat network
- [x] Tested all the test cases.


utils/mocks/abi_utils.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by mockery v2.14.0. DO NOT EDIT.
+// Code generated by mockery v2.9.4. DO NOT EDIT.
```
Yashk767, line 1: All mockery generated files in utils pkg have changed their version 
It's not affecting the tests but should we update mockery to the latest version? so that generated mocks will be in sync for all of us.
@ashish10677.

Yashk767 (changes requested):

utils/mocks/abi_utils.go
ashish10677: Hmmm, for this we'll have to update the go.mod file right? @Yashk767

Yashk767: Yes we can do that.
Its not required for the build though, we just need to have common version for dev purpose.  
We can just decide it externally too.

ashish10677: Fixed @Yashk767

shekhar2807 (approved):

PR #951, opened 2022-09-27
Hotfix-getDataFromAPI

### LINKED ISSUE #950: Reduce retry mechanism to make sure it doesn't exceed state length
Retry twice with a break of 2 sec and don't retry more than that.

Yashk767:
# Description

- Reduced num of retry attempts to get data from API
- Removed redundant retry attempts
- Logged status code and API when status code from response !=200

Fixes #950

### DIFF
--- a/utils/api.go
+++ b/utils/api.go
@@ -3,7 +3,6 @@ package utils
 import (
 	"errors"
 	"net/http"
-	"razor/core"
 	"time"
 
 	"github.com/PaesslerAG/jsonpath"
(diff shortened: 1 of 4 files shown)


utils/api.go
```diff
@@ -24,14 +23,15 @@ func (*UtilsStruct) GetDataFromAPI(url string) ([]byte, error) {
...
+				log.Errorf("API: %s responded with status code %d", url, response.StatusCode)
 				return errors.New("unable to reach API")
 			}
 			body, err = IOInterface.ReadAll(response.Body)
 			if err != nil {
 				return err
 			}
 			return nil
-		}, RetryInterface.RetryAttempts(core.MaxRetries))
+		}, retry.Attempts(2), retry.Delay(time.Second*2))
```
ashish10677, line 34: Can you change core.MaxRetries to 2 and then use core.MaxRetries itself in retry.Attempts?

ashish10677 (changes requested):

utils/api.go
Yashk767: But we are using maxRetries at number of places like contract calls, client functions ,getters setters both. So it would be reflected everywhere. Would that be okay?

ashish10677 (approved):

PR #975, opened 2022-10-12
Added a check if config key is present in yaml file before fetching value

### LINKED ISSUE #970: Value of the config key is considered as 0 if that key is not present in yaml file.
### Description

As we use `viper.Get(key)` to get value associated to key in yaml file which follows - 
```` 
viper.Get function will return a zero value if the key is not found in yaml file. 
````

### Solution
Set the value of the config key to default if that key is not present in yaml file.
To check if a given key exists, the `viper.IsSet()` method has been provided.

### PR DESCRIPTION
**Pull request by Yashk767**
# Description

Added a check if config key is present in yaml file before fetching using `viper.IsSet(KEY)`. If present than fetch the respective value and if not than fetch the default value.

Fixes #970

### DIFF
--- a/cmd/config-utils.go
+++ b/cmd/config-utils.go
@@ -3,6 +3,7 @@ package cmd
 
 import (
 	"github.com/spf13/viper"
+	"razor/core"
 	"razor/core/types"
 	"razor/utils"
 	"strings"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: cmd/config-utils.go [hunk 1]
```diff
@@ -111,7 +127,12 @@ func (*UtilsStruct) GetWaitTime() (int32, error) {
 		return 3, err
```
ashish10677, line 127: Change this value to 1 and keep the default values consistent.

ashish10677 (changes requested):

cmd/config-utils.go
Yashk767: Done @ashish10677

PR #1016, opened 2022-11-07
Cached API Response

### LINKED ISSUE #976: Cache multiple results from same API
- [ ] Add a caching layer for every URL requested. 
- [ ] Add cache expiry = state length

Yashk767:
# Description

Cached API response using in memory caching. Implemented it using map with expiry period, where URL as key and response result as value.

Fixes #976

### DIFF
--- a/utils/api_test.go
+++ b/utils/api_test.go
@@ -3,6 +3,7 @@ package utils
 import (
 	"errors"
 	"github.com/stretchr/testify/mock"
+	"razor/cache"
 	"razor/utils/mocks"
 	"reflect"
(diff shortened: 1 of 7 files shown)


utils/api_test.go
```diff
@@ -110,6 +112,10 @@ func TestGetDataFromAPI(t *testing.T) {
 	}
 }
 
+func TestAPI(t *testing.T) {
```
Yashk767, line 115: Remove this

ashish10677 (approved):
lgtm

PR #1048, opened 2023-01-10
Fetched API key from env file

### LINKED ISSUE #1002: Fetch api keys, secrets etc from env and apply it to jobs
Analysis: some URLs may require authentication or API keys.

Possible solution: Consider reading the keys from env and substituting it while fetching a job.

Yashk767:
# Description
- Headers can be added to GET requests now.
- Updated content-type field of `datasourceURLstruct` to header field.
- Now header field will contain content-type as sub field.
- All supported fields can be added inside header field.
- Also API_Key field can be added to header, the value of the same can be fetched from env file.
 
Fixes #1002 #1053


Yashk767:
While creating a job which requires a API key, we need to follow a pattern which is defined in constants.go.

e.g. if this is the required format for this API
```
/v1/exchangerate/BTC?apikey=73034021746739393292922SAMPLE-KEY
```

Than we can create a job following this,
```
/v1/exchangerate/BTC?apikey={{API_KEY}} 
```

You can the definition here,
https://github.com/razor-network/oracle-node/pull/1048/files#diff-d5ab1755a1badc517f3a8eda33d72004f1aaca97998a7e2a24580f1c466fb1d2R55

is this fine @ashish10677 @SkandaBhat ?

If this is fine then we can convert the PR to ready for review.

Yashk767:
In env file, 
jobName_API_KEY should be used as a keyword.

e.g. If the jobName which requires API key is `eth_Job`,
then the env file should have 
```
eth_Job_API_KEY = 6fa741de1bdd1d91830ba
```

utils/asset.go
```diff
@@ -507,3 +531,29 @@ func (*UtilsStruct) HandleOfficialJobsFromJSONFile(client *ethclient.Client, col
...
+
+	log.Info("Loading env file...")
+	envFileMap, err := godotenv.Read(envFilePath)
+	if err != nil {
+		log.Error("Error in getting env file map: ", err)
+		return "", "", err
+	}
+	log.Debugf("GetKeyWordAndAPIKeyFromENVFile: ENV file map: %v", envFileMap)
+	for keyword, APIKey := range envFileMap {
+		keywordWithDollar := `$` + keyword
```
Yashk767, line 551: Use this from contstants.go

utils/asset.go
```diff
@@ -292,11 +293,29 @@ func (*UtilsStruct) GetDataToCommitFromJob(job bindings.StructsJob, localCache *
...
+			return nil, err
+		}
+		if isAPIKeyRequired {
+			keyword, APIKey, err := GetKeyWordAndAPIKeyFromENVFile(job.Url)
+			if err != nil {
+				log.Error("Error in getting value from env file: ", err)
+				return nil, err
+			}
+			log.Debug("API key: ", APIKey)
+			keywordWithDollar := `$` + keyword
```
Yashk767, line 308: Pick this up from constants.go

Yashk767:
While defining the job which requires a API key, you need to start the keyword with regex defined in constants.go, which is set to `$`. 

e.g. if this is the required format for this API
```
/v1/exchangerate/BTC?apikey=73034021746739393292922
```

Than we can create a job following this,
```
/v1/exchangerate/BTC?apikey=$COINAPI_API_KEY
```

 Now you will youe the same keyword defined after $ in env file.
In env file,
```
COINAPI_API_KEY = 73034021746739393292922
```

utils/api.go
```diff
@@ -27,16 +28,26 @@ func (*UtilsStruct) GetDataFromAPI(dataSourceURLStruct types.DataSourceURL, loca
 		if dataSourceURLStruct.Type == "GET" {
 			err := retry.Do(
 				func() error {
-					response, err := client.Get(dataSourceURLStruct.URL)
+					request, err := http.NewRequest("GET", dataSourceURLStruct.URL, nil)
```
ashish10677, line 31: Move the processing part after the if inside a new function.
It should be something like this:

```
switch(dataSourceURLStruct.Type) {
    case "GET":
            processGetRequest()
    case "POST": 
           processPostRequest()
    default:
          return error
}
```

ashish10677 (changes requested):

ashish10677 (approved):

yohanelly95:
in alpha release

Yashk767:
As suggested
```
// API key should be directly fetched from os environment 
// Use os.Getenv() as used here https://gobyexample.com/environment-variables instead of depending on godotenv library.
// Followed format {API_KEY_KEYWORD} as regex in urls to search for keyword and respective value from os env variables. 
```
Implemented the same in https://github.com/razor-network/oracle-node/pull/1097

While defining the job which requires a API key, staker needs to follow  regex defined in constants.go , which is set to e.g `${API_KEY_KEYWORD}`. 

e.g. if this is the required format for this API
```
https://api.gemini.com/v1/pubticker/v1/exchangerate/BTC?apikey=73034021746739393292922
```

Than a job with url can be created as shown below,
```
https://api.gemini.com/v1/pubticker/v1/exchangerate/BTC?apikey=$COINAPI_API_KEY
```

 Now staker needs to use the same keyword defined inside ${...} as a environment variable and assigning it a value as users API key as shown below.

```
COINAPI_API_KEY = 73034021746739393292922
```

Oracle node will directly look for the same in os environment variables and pick the respective value.
@ashish10677 @SkandaBhat

PR #1058, opened 2023-03-01
Added support for jobs returning hex values

### LINKED ISSUE #1057: Support jobs that return hex values
#### Description

Many POST request return the price feeds in hex values, so oracle node should identify whether the values returned is a hex value and if it is then it should convert into decimal.

Yashk767:
# Description

- Added a new field `returnType` in dataSourceUrl struct which will support jobs returning values other than decimal values.
- Job returning hex values will be supported.
- Removed method caller for functions wherever it was not required to avoid mocking.

Fixes #1057


utils/asset_test.go
```diff
@@ -644,38 +646,28 @@ func TestGetDataToCommitFromJobs(t *testing.T) {
...
-		Url: `{"type": "GET","url": "https://api.gemini.com/v1/pubticker/ethusd","body": {},"header": {}}`,
-	}
-
-	job2 := bindings.StructsJob{Id: 1, SelectorType: 0, Weight: 100,
+	job := bindings.StructsJob{Id: 1, SelectorType: 0, Weight: 100,
 		Power: 2, Name: "ethusd_gemini", Selector: "last",
 		Url: `{"type": "GET","url": "https://api.gemini.com/v1/pubticker/ethusd","body": {},"header": {}}`,
 	}
 
+	//job2 := bindings.StructsJob{Id: 1, SelectorType: 0, Weight: 100,
```
ashish10677, line 654: If this is not needed remove this. If this is needed, uncomment.

ashish10677 (changes requested):

Yashk767:
Requested changes done and replaced alchemy RPC for tests, prior one was failing authentication. @ashish10677

ashish10677 (approved):

Issue #1070, opened 2023-04-06
If RPC is not responding, add support for alternate RPC 


Yashk767:
### Description

If RPC is throwing an error in making contract calls, we should have an alternate RPC that would replace the prior RPC after specific number of retry attempts are done.

PR #1071, opened 2023-04-12
Added a generic function for retry mechanism

### LINKED ISSUE #1004: Implement retry mechanism to all contract calls using a generic function
As every contract and client call is being done using a generic function, so we can add a retry mechanism to the same generic function which will decrease the code repetition.

Yashk767:
# Description

- Added function `InvokeFunctionWithRetryAttempts` which will implement retry mechanism on top of a contract call function.

- Added `alternateProvider` field in config file.
- Added support for alternate RPC which will start working if primary RPC is not responding and will work for `core.SwitchClientDuration` seconds (set to 5 epochs currently) and then node will switch back to primary RPC.

Fixes #1004 #1070


core/constants.go
```diff
@@ -59,3 +60,6 @@ var APIKeyRegex = `$`
 
 var ProcessRequestRetryAttempts uint = 2
 var ProcessRequestRetryDelay = 2
+
+//SwitchClientDuration is the timeout for which alternate client i.e. alternate client will be activated
```
Yashk767, line 64: Change the description comment

client/alternateClient.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/ethereum/go-ethereum/ethclient"
+	"razor/logger"
+	"reflect"
+	"time"
+)
+
+var log = logger.NewLogger()
+
+var SwitchClientToAlternateClient bool
+var AlternateProvider string
```
Yashk767, line 13: Include these 3 fields in a single struct `AlternateClient`

type AlternateClient struct {
log logger
switchToAlternateClient bool
alternateProvider string
}

Write Setter and getter for switchToAlternateClient , alternateProvider

cmd/config-utils.go
```diff
@@ -75,6 +81,8 @@ func (*UtilsStruct) GetConfigData() (types.Configurations, error) {
 		return config, err
 	}
 	config.Provider = provider
+	config.AlternateProvider = alternateProvider
+	client.AlternateProvider = alternateProvider
```
Yashk767, line 85: Set it using setter from `client` package.

client/alternateClient.go
Yashk767: Not added log field in the struct as it had to be called in vote.go for loop and it would have been initialised multiple times. So just kept it same `var log = logger.NewLogger()` in `client` package itself.

ashish10677 (approved):

shekhar2807 (approved):

PR #1092, opened 2023-06-16
Alternate provider fix

### LINKED ISSUE #1091: AlternateProvider should not be used if not set in config
### Description

Right now in `constants.go` on oracle node v1.0.6,
```
var DefaultProvider = "http://127.0.0.1:8545"
var DefaultAlternateProvider = "http://127.0.0.1:8545"
```

If a user doesn’t set `alternateProvider` in config using `setConfig` command, than oracle node will pick up default value from constants.go as alternate provider and try to make contract calls using the same (default value is set to local host which is Invalid RPC on SKALE), so that can create problems when our primary RPC is down.

### Solution

Alternate provider should not be used if it not set in config by user rather than picking up default value from constants.go.

Yashk767:
# Description

Added a check whether `alternateProvder` is not set or equal to empty string before accessing feature of alternate provider.

Fixes #1091

### DIFF
--- a/cmd/config-utils.go
+++ b/cmd/config-utils.go
@@ -133,7 +133,7 @@ func (*UtilsStruct) GetAlternateProvider() (string, error) {
 	}
 	if alternateProvider == "" {
 		if viper.IsSet("alternateProvider") {
-			alternateProvider = viper.GetString("alternatalternateProvider = viper.GetString(\"alternateProvidereProvider")
+			alternateProvider = viper.GetString("alternateProvider")
 		} else {
 			alternateProvider = ""
 			log.Debug("alternate provider is not set, taking its nil value ", alternateProvider)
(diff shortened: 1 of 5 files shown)


cmd/config-utils.go
```diff
@@ -129,14 +129,14 @@ func (*UtilsStruct) GetProvider() (string, error) {
 func (*UtilsStruct) GetAlternateProvider() (string, error) {
 	alternateProvider, err := flagSetUtils.GetRootStringAlternateProvider()
 	if err != nil {
-		return core.DefaultAlternateProvider, err
+		return "", err
 	}
 	if alternateProvider == "" {
 		if viper.IsSet("alternateProvider") {
-			alternateProvider = viper.GetString("alternateProvider")
+			alternateProvider = viper.GetString("alternatalternateProvider = viper.GetString(\"alternateProvidereProvider")
```
ashish10677, line 136: This is a typo

ashish10677 (changes requested):

cmd/config-utils.go
Yashk767: Ohh got refactored wrongly after testing it. Fixing it

Yashk767: Done @ashish10677

ashish10677 (approved):

PR #1113, opened 2023-08-22
Removed default alternate provider from set config script

### LINKED ISSUE #1112: Don't set alternate provider if nil value is provided in config.sh
### Description

Right now in `config.sh` if a staker puts in empty value for alternate provider, then `http://127.0.0.1:8545` gets stored as a value to it. 
If it's a staker on mainnet/testnet than it can face issues when  primary provider gets switched to alternate provider.

So alternate provider should not be set if nil/empty value is provided to it after building from source in binary.

Yashk767:
Fixes #1112

### DIFF
--- a/config.sh
+++ b/config.sh
@@ -11,11 +11,7 @@ then
   PROVIDER="http://127.0.0.1:8545"
 fi
 
-read -rp "Alternate Provider: (http://127.0.0.1:8545) " ALTERNATE_PROVIDER
-if [ -z "$ALTERNATE_PROVIDER" ];
-then
-  ALTERNATE_PROVIDER="http://127.0.0.1:8545"
-fi
+read -rp "Alternate Provider: " ALTERNATE_PROVIDER
 
 read -rp "Gas Multiplier: (1.0) " GAS_MULTIPLIER
 if [ -z "$GAS_MULTIPLIER" ];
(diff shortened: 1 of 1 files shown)


ashish10677 (approved):

ashish10677:
If the user already has existing alternate provider set as 127.0.0.1 from the previous build, then regardless of this change, it'll still be there right? Unless, it is manually removed.

Yashk767:
> If the user already has existing alternate provider set as 127.0.0.1 from the previous build, then regardless of this change, it'll still be there right? Unless, it is manually removed.

Yes it will be still there if it was already set before in the config. 
Should we delete the existing `razor.yaml` file and update it according the latest build when staker tries to build from scratch? This is the only way we can solve the above problem.

ashish10677:
> > If the user already has existing alternate provider set as 127.0.0.1 from the previous build, then regardless of this change, it'll still be there right? Unless, it is manually removed.
> 
> Yes it will be still there if it was already set before in the config. Should we delete the existing `razor.yaml` file and update it according the latest build when staker tries to build from scratch? This is the only way we can solve the above problem, unless staker manually does it.

I think it would be a very radical approach. Let's do this, update the README for this release and make sure the message is sent that if the alternate provider is 127.0.0.1 then please remove the yaml file and rebuild from scratch.

PR #1114, opened 2023-08-30
Updated version to v1.0.6-patch.1.alpha.2

Yashk767:
# Description

Updated version to v1.0.6-patch.1-alpha.1 for next staging release.
This is a part of our next public release `v1.0.6-patch.1`

### DIFF
--- a/core/version.go
+++ b/core/version.go
@@ -6,7 +6,7 @@ const (
 	VersionMajor = 1                 // Major version component of the current release
 	VersionMinor = 0                 // Minor version component of the current release
 	VersionPatch = 6                 // Patch version component of the current release
-	VersionMeta  = "patch.1-alpha.1" // Version metadata to append to the version string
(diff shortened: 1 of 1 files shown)


ashish10677:
Semantic versioning suggests only one hyphen, v1.0.6-patch.1-alpha.1 isn't a correct version:
https://semver.org/#spec-item-9

yohanelly95:
IMO it should be `1.0.6-patch.1.alpha.2` using a single hyphen and if we keep the patch the same.

Yashk767:
> IMO it should be `1.0.6-patch.1.alpha.2` using a single hyphen and if we keep the patch the same.

Yes, I updated the version to same: `1.0.6-patch.1.alpha.2`

ashish10677 (approved):

yohanelly95 (approved):

PR #1154, opened 2023-10-13
Feature/v1.1.0

SkandaBhat:
# Description

This PR merges `v1.1.0` release into `develop` branch.

# Release changes

- Added support for fetching API keys from the environment for requests.
- Introduced support for POST jobs.
- Jobs can now return hex values.
- Version number is now included in logs.
- Updated to the latest versions of `go-ethereum`, `go`, and other dependencies.
- Resolved all major Dependabot alerts by updating and installing node dependencies with the latest node and npm versions.
- Modified the default value and functionality of `gasLimitOverride`. It will now be used every time there's a gas estimation failure.
- Commit data will now be stored even if the transaction fails.
- Nil values for the Alternate provider in the config are no longer supported.


cmd/commit.go
```diff
@@ -145,3 +146,67 @@ func (*UtilsStruct) Commit(client *ethclient.Client, config types.Configurations
 	log.Info("Txn Hash: ", txnHash.Hex())
 	return txnHash, nil
 }
+
+func CalculateSeed(client *ethclient.Client, account types.Account, keystorePath string, epoch uint32) ([]byte, error) {
```
SkandaBhat, line 150: Why was this change made? What exactly changed here?

Yashk767: As we are storing commit data before checking its transaction status, we had to add an additional verification while fetching data from commit data file. So to verify it was to seed calculation, so to avoid redundant code, seed calculation was separated to separate function.
Commit flow - https://www.notion.so/razorto/Oracle-Node-Alpha-Test-Suite-2a7128a3778b4332866931ee1a657ea4?pvs=4#f7b6e30cc46149faba90aa7be63d63e7

config.sh
```diff
@@ -58,4 +54,10 @@ read -rp "Log Files Max Age: (365) " MAX_AGE
 if [ -z "$MAX_AGE" ]; then
    MAX_AGE=365
 fi
-$RAZOR setConfig -p $PROVIDER --alternateProvider $ALTERNATE_PROVIDER -b $BUFFER -g $GAS_MULTIPLIER -w $WAIT_TIME --gasprice $GAS_PRICE --gasLimit $GAS_LIMIT --rpcTimeout 10 --httpTimeout 10 --logFileMaxSize $MAX_SIZE --logFileMaxBackups $MAX_BACKUPS --logFileMaxAge $MAX_AGE
+
+if [ -z "$ALTERNATE_PROVIDER" ]; then
```
SkandaBhat, line 58: Please follow DRY.

This can be rewritten:

```if [ -z "$ALTERNATE_PROVIDER" ]; then
   $RAZOR setConfig -p $PROVIDER -b $BUFFER -g $GAS_MULTIPLIER -w $WAIT_TIME --gasprice $GAS_PRICE --gasLimit $GAS_LIMIT --rpcTimeout 10 --httpTimeout 10 --logFileMaxSize $MAX_SIZE --logFileMaxBackups $MAX_BACKUPS --logFileMaxAge $MAX_AGE
 else
 $RAZOR setConfig -p $PROVIDER --alternateProvider "$ALTERNATE_PROVIDER" -b $BUFFER -g $GAS_MULTIPLIER -w $WAIT_TIME --gasprice $GAS_PRICE --gasLimit $GAS_LIMIT --rpcTimeout 10 --httpTimeout 10 --logFileMaxSize $MAX_SIZE --logFileMaxBackups $MAX_BACKUPS --logFileMaxAge $MAX_AGE
```

core/constants.go
```diff
@@ -10,7 +10,7 @@ import (
 
 var EpochLength uint64 = 1200
 var NumberOfStates uint64 = 5
-var ChainId = big.NewInt(0x109B4597)
+var ChainId = big.NewInt(0x5a79c44e)
```
SkandaBhat, line 13: Please add a comment here which says which chainId is being used.

utils/api.go
```diff
@@ -1,53 +1,77 @@
...
-				body, err = IOInterface.ReadAll(response.Body)
-				if err != nil {
-					return err
-				}
-				return nil
-			}, retry.Attempts(2), retry.Delay(time.Second*2))
-		if err != nil {
-			return nil, err
+	cachedData, cachedErr := localCache.Read(dataSourceURLStruct.URL)
+	if cachedErr != nil {
```
SkandaBhat, line 26: There is a lot of code repitition in this block. Please rewrite so that its easy to understand and follows DRY.

SkandaBhat, line 26: Its more idiomatic in Go to check for ```cachedErr == nil``` and return early, reducing the level of indentation for the main logic.

utils/asset.go
```diff
@@ -483,3 +520,21 @@ func (*UtilsStruct) HandleOfficialJobsFromJSONFile(client *ethclient.Client, col
 
 	return overrideJobs, overriddenJobIds
 }
+
+func ReplaceValueWithDataFromENVFile(re *regexp.Regexp, value string) string {
```
SkandaBhat, line 524: Can be rewritten to be much cleaner:

```func ReplaceWithEnvValue(re *regexp.Regexp, value string) string {
    substrings := re.FindAllString(value, -1)
    replacedKeywords := make(map[string]bool)

    for _, keyword := range substrings {
        if _, replaced := replacedKeywords[keyword]; !replaced {
            envValue, exists := os.LookupEnv(keyword)
            if exists {
                value = strings.Replace(value, keyword, envValue, -1)
                replacedKeywords[keyword] = true
            }
            // Handle the case where the environment variable doesn't exist, if necessary
        }
    }
    return value
}
```

utils/asset.go
```diff
@@ -483,3 +520,21 @@ func (*UtilsStruct) HandleOfficialJobsFromJSONFile(client *ethclient.Client, col
...
+	// substrings denotes all the occurrences of substring which satisfies APIKeyRegex
+	substrings := re.FindAllString(value, -1)
+	log.Debug("ReplaceValueWithDataFromENVFile: Substrings array: ", substrings)
+	// Fetching keyword and respective value from env file
+	for i := 0; i < len(substrings); i++ {
```
SkandaBhat, line 529: Why is for loop written like this? Should be

```
for _, keyword := range substrings {
```

utils/asset.go
```diff
@@ -272,35 +273,60 @@ func (*UtilsStruct) GetDataToCommitFromJobs(jobs []bindings.StructsJob, localCac
...
 	var parsedJSON map[string]interface{}
 	var (
-		response []byte
-		apiErr   error
+		response            []byte
+		apiErr              error
+		dataSourceURLStruct types.DataSourceURL
 	)
+	log.Debugf("Getting the data to commit for job %s having job Id %d", job.Name, job.Id)
+	if strings.HasPrefix(job.Url, "{") {
```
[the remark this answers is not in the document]
Yashk767: We support both source: `job.url` being passed as direct url string which can be used for direct GET jobs and
source being passed as JSON formatted string, this was mainly added for POST requests but can be used or POST jobs and GET jobs passed in the form of struct
```
"source": {
      Type       string                 `json:"type"`
	URL        string                 `json:"url"`
	Body       map[string]interface{} `json:"body"`
	Header     map[string]string      `json:"header"`
	ReturnType string                 `json:"returnType"`
    }
```

so ` strings.HasPrefix(job.Url, "{") {` identifies that source is passed as JSON formatted string so it would be treated as job passed as struct and decoded accordingly.

utils/math.go
```diff
@@ -31,6 +34,23 @@ func (*UtilsStruct) ConvertToNumber(num interface{}) (*big.Float, error) {
 	return big.NewFloat(0), nil
 }
 
+func ManageReturnType(num interface{}, returnType string) (interface{}, error) {
```
SkandaBhat, line 37: Function name is super vague. Rename it to something more specific for eg:

```
ConvertToReturnType
```

```
TransformToReturnType
```

SkandaBhat: Seems to be just converting hex to int.

SkandaBhat, line 37: This entire function can be simplified:

```
func ConvertHexToInt(hexValue string) (int64, error) {
    // Removing "0x" prefix from hex value, if present
    hexValue = strings.TrimPrefix(hexValue, "0x")
    
    // Converting the given hex value to its decimal representation
    decimalValue, err := strconv.ParseInt(hexValue, 16, 64)
    if err != nil {
        log.Errorf("Error converting hex value %v to integer: %v", hexValue, err)
        return 0, err
    }
    
    return decimalValue, nil
}
```

utils/api.go
```diff
@@ -56,15 +80,55 @@ func (*UtilsStruct) GetDataFromJSON(jsonObject map[string]interface{}, selector
...
 	})
-	err := c.Visit(url)
+	err := c.Visit(dataSourceURLStruct.URL)
 	if err != nil {
 		return "", err
 	}
 	return priceData, nil
 }
+
+func AddHeaderToRequest(request *http.Request, headerMap map[string]string) (*http.Request, error) {
```
SkandaBhat, line 96: This function seems to have multiple responsibilities:
- checking if a header value matches an API key pattern and replacing it with an environment variable
- adding headers to an HTTP request.

But the name `AddHeaderToRequest` suggests that it does only one thing. Move out the different logics into different functions and refactor.

utils/api.go
```diff
@@ -56,15 +80,55 @@ func (*UtilsStruct) GetDataFromJSON(jsonObject map[string]interface{}, selector
...
+	for key, value := range headerMap {
+		re := regexp.MustCompile(core.APIKeyRegex)
```
SkandaBhat, line 98: why is this done inside the loop? we are calculating again and again when this can be done only once outside the loop.

utils/api.go
```diff
@@ -56,15 +80,55 @@ func (*UtilsStruct) GetDataFromJSON(jsonObject map[string]interface{}, selector
...
+		isAPIKeyRequired := re.MatchString(value)
+		if isAPIKeyRequired {
+			value = ReplaceValueWithDataFromENVFile(re, value)
+		}
+		log.Debugf("Adding key: %s, value: %s pair to header", key, value)
+		request.Header.Add(key, value)
+	}
+	return request, nil
```
SkandaBhat, line 106: are there no error scenarios for the function?

utils/api.go
```diff
@@ -56,15 +80,55 @@ func (*UtilsStruct) GetDataFromJSON(jsonObject map[string]interface{}, selector
...
+}
+
+func ProcessRequest(client http.Client, dataSourceURLStruct types.DataSourceURL, requestBody io.Reader) ([]byte, error) {
+	request, err := http.NewRequest(dataSourceURLStruct.Type, dataSourceURLStruct.URL, requestBody)
```
SkandaBhat, line 110: dataSourceURLStruct.Type can be anything. Needs to be validated before creating a new request.

utils/api.go
```diff
@@ -56,15 +80,55 @@ func (*UtilsStruct) GetDataFromJSON(jsonObject map[string]interface{}, selector
...
+		log.Errorf("Error in adding header to %s request: %v", dataSourceURLStruct.Type, err)
+		return nil, err
+	}
+	response, err := client.Do(requestWithHeader)
+	if err != nil {
+		log.Errorf("Error sending %s request URL %s: %v", dataSourceURLStruct.Type, dataSourceURLStruct.URL, err)
+		return nil, err
+	}
+	defer response.Body.Close()
+	if response.StatusCode != 200 {
```
SkandaBhat, line 125: there can be many status code responses that are other than 200 that are successful. 

https://stackoverflow.com/a/55210335

utils/api.go
```diff
@@ -56,15 +80,55 @@ func (*UtilsStruct) GetDataFromJSON(jsonObject map[string]interface{}, selector
...
+		log.Errorf("API: %s responded with status code %d", dataSourceURLStruct.URL, response.StatusCode)
+		return nil, errors.New("unable to reach API")
```
SkandaBhat, line 127: "unable to reach API" is misleading. Rewrite to something more appropriate.

SkandaBhat (commented):
There are a lot of code quality improvements that are suggested in this PR. Please address all of them.

utils/math.go
Yashk767: This was being done in order to be open to more return types in future and all of them will be handled in the same function. Right now we support only hex thats why just kept hex the only case, Should I still rename?

utils/api.go
Yashk767: Yes makes sense, will move it outside.

utils/asset.go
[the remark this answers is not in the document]
SkandaBhat: There has to be a better way to identify this. cc: @ashish10677

utils/math.go
SkandaBhat: There is no point keeping something customisable for the future. If it needs to be customisable, then it needs to be customisable right now.

The requirement for customisability seems very vague and open ended, and probably wont even warrant an updation in the future.

If you think this will be customised, then add all the customisations right now. In go, each function needs to have one responsibility and nothing else.

utils/api.go
Yashk767: Yes there are no error cases for this, this function adds header to the request if present else return the same request. Makes sense, returning error can be removed from this function. Noted.

utils/api.go
Yashk767: yes we do this inside `GetDataFromAPI()` where we call `ProcessRequest()` only for cases when  `dataSourceURLStruct.Type` is "GET" or "POST", else we return invalid request type.

utils/asset.go
Yashk767: yes this can be written like this as well. Suggested one is a good practise, will replace.

utils/asset.go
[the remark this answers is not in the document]
ashish10677:
```
func isJSONCompatible(s string) bool {
    var temp interface{}
    err := json.Unmarshal([]byte(s), &temp)
    return err == nil
}
```

You can use this function to know whether the string is a JSON compatible string or a normal string. Use this instead of doing strings.HasPrefix().

utils/api.go
ashish10677:
```
// Success is indicated with 2xx status codes:
statusOK := response.StatusCode >= 200 && response.StatusCode < 300
if !statusOK {
    fmt.Println("Non-OK HTTP status:", response.StatusCode)
    // You may read / inspect response body
    return
}
```
You can use this @Yashk767

utils/api.go
ashish10677:
```
return nil, fmt.Errorf("HTTP request failed with status code %d: %s", response.StatusCode, response.Status)
```
This can be used.

ashish10677:
> @Yashk767 coverage seems to be decreasing here, any reason for it?

@Yashk767 add tests for the new functions.

cmd/commit.go
```diff
@@ -145,3 +146,67 @@ func (*UtilsStruct) Commit(client *ethclient.Client, config types.Configurations
...
+	salt, err := cmdUtils.GetSalt(client, epoch)
+	if err != nil {
+		log.Error("Error in getting salt: ", err)
+		return nil, err
+	}
+	seed := solsha3.SoliditySHA3([]string{"bytes32", "bytes32"}, []interface{}{"0x" + hex.EncodeToString(salt[:]), "0x" + hex.EncodeToString(secret)})
+	return seed, nil
+}
+
+func CalculateCommitment(seed []byte, values []*big.Int) ([32]byte, error) {
```
ashish10677, line 166: Add tests for this function and CalculateSeed

ashish10677 (changes requested):

utils/api.go
Yashk767: Yes makes sense - Go idioms favor early returns.

Currently `localCache.Read()` returns custom error `erros.New("data not present in cache")` when no data found in cache and this is the only error being returned else it returns cachedData without any error.

This can be done in a better way - by using a `boolean`  return instead of `error`.
So GetDataFromAPI() will be changed to
```
cachedData, found := localCache.Read(dataSourceURLStruct.URL)
	if found :
		log.Debugf("Getting Data for URL %s from local cache...", dataSourceURLStruct.URL)
		return cachedData, nil
	else :
                makeAPIRequest
 ```
Its more readable and favours early returns as well.

File: utils/asset.go (hunk 7 above)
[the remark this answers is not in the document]
Comment (Yashk767) [reply utils/asset.go:281]: This `isJSONCompatible()` doesn't work in our case as technically a normal string "Hello" is also a valid JSON.

File: utils/math.go (hunk 8 above)
Comment (Yashk767) [reply utils/math.go:37]: Yes I did change the function. 
Here the function should rather look like `ConvertHexToInt(parsedData interface{})` where parameter should be of type `interface{}`as the input for function is parsed data from json which could be of any type.

If we stringify the parsed data and pass it as parameter, it will also convert the non string data type into string and can lead to invalid conversion.
So inside `ConvertHexToInt()` we return error if the data type is not hex string to avoid invalid conversions.

```
func ConvertHexToInt(num interface{}) (int, error) {
	switch v := num.(type) {
	case string:
		hexValue := strings.TrimPrefix(v, "0x")
		decimalValue, err := strconv.ParseInt(hexValue, 16, 64)
		if err != nil {
			log.Errorf("Error converting hex value %v to integer: %v", hexValue, err)
			return 0, err
		}
		return int(decimalValue), nil
	default:
		log.Errorf("Input data %v is not a hex value", num)
		return 0, errors.New("input data is not hex")
	}
}
```

File: utils/asset.go (hunk 7 above)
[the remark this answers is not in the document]
Comment (ashish10677) [reply utils/asset.go:281]: But you're checking for strings beginning with `"{"`. By that logic "Hello" will fail for HasPrefix("{") as well

File: cmd/commit_test.go [hunk 16]
```diff
@@ -406,3 +407,161 @@ func BenchmarkHandleCommitState(b *testing.B) {
...
+		name    string
+		args    args
+		want    bool
+		wantErr bool
+	}{
+		{
+			name: "Test 1: When commitment is verified successfully",
+			args: args{
+				commitment: types.Commitment{
+					CommitmentHash: [32]byte{34, 201, 186, 7, 78, 68, 208, 0, 145, 22, 178, 68, 165, 206, 206, 158, 154, 222, 133, 175, 72, 110, 31, 79, 141, 184, 227, 4, 230, 96, 91, 234},
```
ashish10677, line 505: Instead of keeping it bytes,  a better practice would be to keep the data in hex, so that it is readable, can convert into bytes wherever required.

ashish10677 (changes requested):

utils/asset.go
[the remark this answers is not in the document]
Yashk767: Hmm just rechecked this, actually `isJSONCompatible()` should work. Let me clearify, 
These are the outputs for `isJSONCompatible(job.Url)` for different `job.Url` cases.

```
CASE 1: "Hello":   is an invalid JSON
CASE 2: `"Hello"`: is a valid JSON
```

The outer quotes are part of the Go string, and the inner double quotes represent the JSON string.
so CASE 1 checks `Hello` as JSON string and return false.
and CASE 2 check `"Hello"` as JSON string which is true and valid as strings in JSON need to be wrapped in double quotes which is true in this case.

In our case of direct Url strings, isJSONCompatible() will work correctly.
```
e.g.

job.Url = "https://api.gemini.com/v1/pubticker/ethusd"
isJSONCompatible(job.Url) returns -> false

File: cmd/commit_test.go (hunk 16 above)
Comment (Yashk767) [reply cmd/commit_test.go:514]: Yes its done in https://github.com/razor-network/oracle-node/pull/1154/commits/6401d80c617e99b9e65926f0ca568cf3720945d8
But it could not be done for `CommitmentHash` as it is a field inside defined `types.Commitment` struct so we cannot change its data type. and we cannot directly use `decodeHexToBytes(CommitmentHexValue)` in arguments as it returns `([]byte, error)`.
Else I have replaced  all the byte values with hex strings to be more readable at all possible places.

Comment (ashish10677) [reply cmd/commit_test.go:514]: You can fix it actually. Suppose args is set up like this:
```
type args struct {
		values        []*big.Int
		commitment    string
		commitmentErr error
		secret        string
		secretErr     error
		salt          string
	}
```
And then, before this line:
```
			utilsMock.On("GetCommitment", mock.Anything, mock.Anything).Return(tt.args.commitment, tt.args.commitmentErr)
```
You convert the commitment string into byte and place it inside the struct.

Comment (Yashk767) [reply cmd/commit_test.go]: Addressed it in https://github.com/razor-network/oracle-node/pull/1154/commits/06ee3fadda8c50eb07f7657a5d9b49a01e632ba3

**Review by ashish10677 — APPROVED** [review]

**Comment by SkandaBhat** [comment]
Were negative test cases added?

```
var apiKeys = []string{
    "key123?break",
    "key&124break",
    "key=125break",
    "key#126break",
    "key+127break",
    "key%128break",
    "key/129break",
    "key?&=#+%/130",
}
```

Please confirm if these do not break the feature. cc: @ashish10677 @Yashk767

**Comment by Yashk767** [comment]
> Were negative test cases added?
> 
> ```
> var apiKeys = []string{
>     "key123?break",
>     "key&124break",
>     "key=125break",
>     "key#126break",
>     "key+127break",
>     "key%128break",
>     "key/129break",
>     "key?&=#+%/130",
> }
> ```
> 
> Please confirm if these do not break the feature. cc: @ashish10677 @Yashk767

Yes I checked this, `ReplaceValueWithDataFromENVFile()` does two things
1. Find keyword that match the regex pattern: 
     - This works fine for all the special characters.
2. For that keyword, the function uses `os.ExpandEnv(keyword)` to fetch its corresponding value from the environment variables:
      - `os.ExpandEnv(keyword)` also works fine as sexpected for all the special characters.
     
NOTE:
But just to note that the cases which contain `=` in their keyword are not valid keywords and can be users fault, as keyword is something which is set by user in environment variables.
And Windows/MacOs uses the equals `=` sign as a delimiter between names and values in its environment variable block
```     
 e.g. 
if you had an environment variable like MY=VAR=SomeValue, the system 
might misinterpret this as an environment variable named MY with the value 
VAR=SomeValue, which is not the intended result  
```

So `os.ExpandEnv()` in our function returns empty string if no keyword found in env variables which is a valid case.

CONCLUSION:
`ReplaceValueWithDataFromENVFile()` works as expected and handles all the possible edge cases mentioned above.

cc @SkandaBhat @ashish10677

PR #1168, opened 2023-12-01
feat: Fetched each job for each collection in a separate go routine

Yashk767:
# Description

Fetching of jobs for a collection is implemented  in parallel.

Fixes https://linear.app/interstellar-research/issue/RAZ-599

# How Has This Been Tested?

Ran a staker on devnet set up, and also modified tests to look for values being reported correctly

### DIFF
--- a/utils/asset.go
+++ b/utils/asset.go
@@ -14,6 +14,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/avast/retry-go"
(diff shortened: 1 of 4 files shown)


utils/asset.go
```diff
@@ -253,21 +254,31 @@ func (*UtilsStruct) GetActiveCollection(client *ethclient.Client, collectionId u
 	return collection, nil
 }
 
-func (*UtilsStruct) GetDataToCommitFromJobs(jobs []bindings.StructsJob, localCache *cache.LocalCache) ([]*big.Int, []uint8, error) {
-	var (
-		data   []*big.Int
-		weight []uint8
-	)
+func (*UtilsStruct) GetDataToCommitFromJobs(jobs []bindings.StructsJob, localCache *cache.LocalCache) ([]*big.Int, []uint8) {
+	var wg sync.WaitGroup
```
ashish10677, line 258: Write this in one bracket:
```
var (
    mu sync.Mutex
    data []*big.Int
   .
   .
   .
)
```

utils/asset.go
```diff
@@ -253,21 +254,31 @@ func (*UtilsStruct) GetActiveCollection(client *ethclient.Client, collectionId u
...
 	for _, job := range jobs {
-		dataToAppend, err := UtilsInterface.GetDataToCommitFromJob(job, localCache)
-		if err != nil {
-			continue
-		}
-		log.Debugf("Job %s gives data %s", job.Url, dataToAppend)
-		data = append(data, dataToAppend)
-		weight = append(weight, job.Weight)
+		wg.Add(1)
+		go func(job bindings.StructsJob) {
```
ashish10677, line 265: move this function out, and call it as `go funcName()`

utils/asset.go
```diff
@@ -253,21 +254,31 @@ func (*UtilsStruct) GetActiveCollection(client *ethclient.Client, collectionId u
...
+			defer wg.Done()
+			dataToAppend, err := UtilsInterface.GetDataToCommitFromJob(job, localCache)
+			if err != nil {
+				return
```
ashish10677, line 269: Why is there no logging of error, or sending error into an error channel?

ashish10677 (changes requested):

utils/asset.go
Yashk767: The error here is logged inside `GetDataToCommitFromJob()` so we don't log here, and we cannot return this error as we need to iterate over all other jobs so we just use `continue` keyword here. 
We don't accumulate these errors and we don't use this error so there was no need to send the job errors anywhere to a channel.

ashish10677: So if there is no error, the default option is to proceed to the next job?

Yashk767: yes

utils/asset.go
Yashk767: Inside go routine we are only calling, `GetDataToCommitFromJobs` with job as a parameter, I don't think we need a separate function just for this right?

utils/asset.go
Yashk767: Done in https://github.com/razor-network/oracle-node/pull/1168/commits/a772e124ec8366c28c11e3fa7c3a32da4cc217fe

utils/asset.go
ashish10677: But it is consuming space in this function, making this function bigger. It's better to move it out.

Yashk767: Done in https://github.com/razor-network/oracle-node/pull/1168/commits/dca6b3335907419a709e11cfbf2c0f7e78bd27ba

ashish10677 (approved):

PR #1174, opened 2023-12-20
feat: added support for job returning results in hex array

Yashk767:
# Description

Added support for job like uniswap v2 datasource result which are of type hex conversion of `uint256[]`.

Fixes https://linear.app/interstellar-research/issue/RAZ-634

# How Has This Been Tested?

Created a uniswap v2 job which returns value in the form of converted hex array ans ran stakers on local to check whether report correct value or not.


ashish10677 (approved):
lgtm

rajkharvar (approved):

PR #1176, opened 2024-01-04
refactor: used hash of url and body as a cache key

Yashk767:
# Description

Primary key for caching is changed to`hash of URL and body` instead of just `URL` as most of the POST  jobs that we have the same URL which is the RPC. 

Fixes https://linear.app/interstellar-research/issue/RAZ-658

# How Has This Been Tested?

Ran a staker on staging when it has multiple post jobs assigned in the same epoch.

### DIFF
--- a/utils/asset_test.go
+++ b/utils/asset_test.go
@@ -600,31 +600,30 @@ func TestGetDataToCommitFromJobs(t *testing.T) {
 	}
 
 	tests := []struct {
-		name            string
-		args            args
-		wantArrayLength int
-		wantErr         bool
+		name    string
+		args    args
+		wantErr bool
 	}{
 		{
 			name: "Test 1: Getting values from set of jobs of length 4",
 			args: args{
 				jobs: jobsArray[:4],
 			},
-			wantArrayLength: 4,
+			wantErr: false,
 		},
 		{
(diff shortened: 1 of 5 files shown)


utils/asset_test.go
```diff
@@ -554,95 +554,95 @@ func TestGetAllCollections(t *testing.T) {
...
-				dataToAppend:       big.NewInt(1),
+				jobs: jobsArray[:2],
 			},
-			want:    []*big.Int{big.NewInt(1), big.NewInt(1)},
-			wantErr: false,
+			wantArrayLength: 2,
 		},
 		{
-			name: "Test 3: When there is an error in getting jobPath",
+			name: "Test 3: Getting values from whole set of jobs of length 9 but job at last index reports an error",
```
ashish10677, line 623: Job 5 returns an error not the last job

ashish10677 (changes requested):

utils/asset_test.go
Yashk767: Changed to index 5 in test name in https://github.com/razor-network/oracle-node/pull/1176/commits/a5906d39aa2ca5e176a29470d7987d3fd71fa258

ashish10677 (approved):

yohanelly95 (approved):

PR #1183, opened 2024-02-12
feat: caching jobs and collections

Yashk767:
# Description

- Stored jobs and collections in a cache and updating the cache on every collection/job update or creation.
- Faster fetching [O(1)] of jobs from cache for every collection/job resulting in quicker execution of commit process.

Fixes https://linear.app/interstellar-research/issue/RAZ-713

# How Has This Been Tested?

Tested with running a staker and creating create and update jobs/collections transactions and checked whether event listeners are listening correctly and updating the cache with the expected update value.


ashish10677 (approved):
lgtm

rajkharvar (approved):

PR #1192, opened 2024-02-26
fix: incorrect state error for shorter epoch length

Yashk767:
# Description

Transactions were missing state due to delayed state actions consuming more time than expected.
- This was due to higher values of config parameters `bufferPercent` and `wait`.
- And resetting of jobs/collections cache was part of main voting process which would consume some time.
 

Fixes 
https://linear.app/interstellar-research/issue/RAZ-755
https://linear.app/interstellar-research/issue/RAZ-758

# How Has This Been Tested?

Ran a staker on staging with updated config values.

### DIFF
--- a/cmd/config-utils.go
+++ b/cmd/config-utils.go
@@ -209,8 +209,8 @@ func (*UtilsStruct) GetMultiplier() (float32, error) {
 //This function returns the buffer percent
 func (*UtilsStruct) GetBufferPercent() (int32, error) {
 	const (
-		MinBufferPercent = 10
-		MaxBufferPercent = 30
+		MinBufferPercent = 0
+		MaxBufferPercent = 5
 	)
 
 	bufferPercent, err := getConfigValue("buffer", "int32", core.DefaultBufferPercent, "buffer")
@@ -238,8 +238,8 @@ func (*UtilsStruct) GetBufferPercent() (int32, error) {
 //This function returns the wait time
 func (*UtilsStruct) GetWaitTime() (int32, error) {
 	const (
-		MinWaitTime = 1  // Minimum wait time in seconds
-		MaxWaitTime = 30 // Maximum wait time in seconds
+		MinWaitTime = 1 // Minimum wait time in seconds
+		MaxWaitTime = 5 // Maximum wait time in seconds
 	)
 
 	waitTime, err := getConfigValue("wait", "int32", core.DefaultWaitTime, "wait")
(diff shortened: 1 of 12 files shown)


cmd/config-utils.go
```diff
@@ -238,8 +238,8 @@ func (*UtilsStruct) GetBufferPercent() (int32, error) {
 //This function returns the wait time
 func (*UtilsStruct) GetWaitTime() (int32, error) {
 	const (
-		MinWaitTime = 1  // Minimum wait time in seconds
-		MaxWaitTime = 30 // Maximum wait time in seconds
+		MinWaitTime = 1 // Minimum wait time in seconds
+		MaxWaitTime = 5 // Maximum wait time in seconds
```
ashish10677, line 242: Are you sure 5 secs will be enough? Blockchain will give block mining confirmation in 5 secs?

cmd/vote.go
```diff
@@ -132,16 +132,8 @@ func (*UtilsStruct) Vote(ctx context.Context, config types.Configurations, clien
...
-			if err := utils.InitJobsCache(client); err != nil {
-				log.Error("Error in initializing jobs cache: ", err)
-				continue
-			}
-			if err := utils.InitCollectionsCache(client); err != nil {
-				log.Error("Error in initializing collections cache: ", err)
-				continue
-			}
+			log.Info("ASSET CACHE EXPIRED! INITIALIZING JOBS AND COLLECTIONS CACHE AGAIN...")
+			go utils.ResetAssetCache(client)
```
ashish10677, line 136: When does this goroutine stop. Not using wait groups isn't a good thing.

ashish10677 (changes requested):

cmd/config-utils.go
Yashk767: No no , this isn't `waitForBlockCompletion` time , this is wait time from the config.
we use this in the voting process during buffer state.

cmd/vote.go
Yashk767: In `ResetAssetCache` we just reset cache where we delete the jobs/collections cache and initialise it again, after resetting this, go routine stops.

But yes I can add a validation using wait groups to make sure the go routine completes its execution, if not than report the error.

cmd/config-utils.go
ashish10677: Ok then fine.

cmd/vote.go
ashish10677: Yes, please do. Without it this may cause uncertainities.
Do check the number of goroutines before and after closing this one, just to confirm, while testing.

Yashk767: Returned error when occurred in resetAssetCache go routine in https://github.com/razor-network/oracle-node/pull/1192/commits/d9eaa5f45799d5aa7f1abdd15ca24064a907e3fa

rajkharvar (dismissed):

cmd/vote.go
Yashk767: As discussed, introduced go routine at the start which will handle resetting cache part - https://github.com/razor-network/oracle-node/pull/1192/commits/a87b16f5858e6d321fb292a7485ff5becfdb2dc4 @ashish10677

ashish10677 (approved):

rajkharvar (approved):

PR #1194, opened 2024-03-19
refactor: used a common http client for all the requests 

### LINKED ISSUE #1197: Number of goroutines were increasing gradually
# Problem

With every iteration in infinite voting for loop, we see a gradual increase in number of goroutines for few epochs and the count does go down after certain epochs but goes on gradual increase which can cause over usage if CPU if the node is running for a long time.

Check whether every goroutine created should be ensured that its getting closed after its execution.

Check for go routines in a for loop are handled correctly.

Check whether every channel is closed using a defer keyword.

# Causes 
1. Currently node creates separate http client instance for every API and makes request using `client.Do()` which creates a new goroutine on every call 

- Using a common http client for all the requests will help here as the new request can use the existing connections from the connection pool and there will be no need create a new connection everytime in a separate go routine.

2.  `keystore.Accounts()` opens a new goroutine everytime whenever we try to fetch keystore files present in keystore directory. - https://github.com/razor-network/oracle-node/issues/1195

Yashk767:
# Description

Used a common http client to avoid opening a new connection on every `client.Do()` call for every API.
A single http client makes the node to reuse the existing connections from connection pool.

Fixes #1197 ,  https://linear.app/interstellar-research/issue/RAZ-771

# How Has This Been Tested?

Ran a node on local and staging environment with 5 min epoch time.

### DIFF
--- a/client/httpClient.go
+++ b/client/httpClient.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"net/http"
+	"razor/core"
 	"time"
 )
 
@@ -11,8 +12,8 @@ func InitHttpClient(httpTimeout int64) {
 	httpClient = &http.Client{
 		Timeout: time.Duration(httpTimeout) * time.Second,
 		Transport: &http.Transport{
-			MaxIdleConns:        5,
-			MaxIdleConnsPerHost: 5,
+			MaxIdleConns:        core.HTTPClientMaxIdleConns,
+			MaxIdleConnsPerHost: core.HTTPClientMaxIdleConnsPerHost,
 		},
 	}
 }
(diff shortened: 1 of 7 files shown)


ashish10677:
@Yashk767 why are the tests failing?

ashish10677 (approved):

client/httpClient.go
```diff
@@ -0,0 +1,23 @@
+package client
+
+import (
+	"net/http"
+	"razor/core"
+	"time"
+)
+
+var httpClient *http.Client
```
SkandaBhat, line 9: DO not use global variables. Use a struct and implement methods. cc: @ashish10677

SkandaBhat (changes requested):

SkandaBhat:
Closing this. Please reopen after updating.

Issue #1195, opened 2024-03-20
Fetching private key every time increases goroutine count


Yashk767:
### Problem

We use GetPrivateKeyFromKeystore(address, password, keystorePath) to fetch the private key from the keystore.

```
func (AccountUtils) GetPrivateKey(address string, password string, keystorePath string) (*ecdsa.PrivateKey, error) {
	allAccounts := AccountUtilsInterface.Accounts(keystorePath)
	for _, account := range allAccounts {
		if strings.EqualFold(account.Address.Hex(), address) {
			return AccountUtilsInterface.GetPrivateKeyFromKeystore(account.URL.Path, password)
		}
	}
	return nil, errors.New("no keystore file found")
}
```

where we iterate over all the accounts keystore files using ks.Accounts and try to match the desired keystore file for the node. 

So here the KeyStore uses file watching mechanisms to automatically detect changes in the keystore directory (adding or removing key files). This involves goroutines waiting for filesystem events.

So when we do ks.Accounnts every time it creates a new goroutine which keep waiting infinitely.

### Solution

Either keep reloading ks.Accounts() in a separate go routine in a separate cache and add a getter function in GetPrivateKeyFromKeystore() which will just get the whole list of accounts keystore files.  [Check security measures]

 Find a way to skip using ks.Accounts call to fetch private key. 

 Store private key in memory at the start and reuse it when needed. [Check security measures]

PR #1196, opened 2024-03-20
refactor: optimised fetching key from keystore dir

Yashk767:
# Description

Previously, when fetching keys from the keystore, we used ks.Accounts(), which created a new goroutine every time. This approach led to goroutines waiting indefinitely, which was inefficient and could potentially cause resource exhaustion issues. To optimize this process and avoid unnecessary resource consumption, the solution is to use the same keystore instance every time to call ks.Accounts() will solves the issue of increasing go routines.

So this PR optimises keystore file fetching to prevent indefinite waiting by goroutines.

Fixes https://github.com/razor-network/oracle-node/issues/1195

Solution
Creating and initialising the keystore instance once at the start and using the same to call `ks.Accounts()` at every fetching of pKey call doesnt increase the goroutine count and it also gives the updated keystore cache at the time of fetching accounts. 

How Has This Been Tested?
- Added realistic fetching of pKey from keystore file tests.
- Account Imports: Tested the creation and importation of both new and old accounts to verify compatibility and functionality across different account types and ages.


ashish10677:
Tests failing @Yashk767

accounts/accounts.go
```diff
@@ -62,3 +68,44 @@ func (AccountUtils) SignData(hash []byte, account types.Account, defaultPath str
 	}
 	return AccountUtilsInterface.Sign(hash, privateKey)
 }
+
+// FindKeystoreFileForAddress matches the keystore file for the given address.
+func FindKeystoreFileForAddress(keystoreDirPath, address string) (string, error) {
+	normalizedAddress := strings.ToLower(address)
+	if strings.HasPrefix(normalizedAddress, "0x") {
+		normalizedAddress = normalizedAddress[2:]
```
ashish10677, line 76: @Yashk767 Please check this suggestion. It's the right way.

ashish10677 (changes requested):

accounts/accounts.go
Yashk767: yes, some tests are failing, fixing those. Will push it together.

.github/workflows/ci.yml
```diff
@@ -60,7 +60,7 @@ jobs:
 
       - name: Execute test case
         run: >
-          go-acc ./... --ignore razor/accounts/mocks --ignore razor/cmd/mocks
+          go-acc ./... --ignore razor/accounts/mocks --ignore razor/accounts/test_accounts --ignore razor/cmd/mocks
```
ashish10677, line 63: Why have you excluded `test_accounts`?

Yashk767: `test_accounts` directory has keystore files of testing addresses

ashish10677 (dismissed):

SkandaBhat:
`package accounts

import (
	"crypto/ecdsa"
	"github.com/ethereum/go-ethereum/accounts"
	"github.com/ethereum/go-ethereum/accounts/keystore"
	"razor/logger"
	"razor/path"
	"os"
	"strings"
)

type AccountManager struct {
	log *logger.Logger
	ks  *keystore.KeyStore
}

// NewAccountManager creates a new AccountManager instance.
func NewAccountManager() *AccountManager {
	return &AccountManager{
		log: logger.NewLogger(),
	}
}

// InitializeKeystore initializes the keystore instance.
func (am *AccountManager) InitializeKeystore(keystorePath string) {
	am.log.Info("Initialising keystoreInstance...")
	am.ks = keystore.NewKeyStore(keystorePath, keystore.StandardScryptN, keystore.StandardScryptP)
}

// CreateAccount creates a new account.
func (am *AccountManager) CreateAccount(keystorePath string, password string) accounts.Account {
	// Method implementation...
}

// GetPrivateKey retrieves the private key from the keystore.
func (am *AccountManager) GetPrivateKey(address string, password string, keystoreDirPath string) (*ecdsa.PrivateKey, error) {
	// Method implementation...
}

// SignData signs the data.
func (am *AccountManager) SignData(hash []byte, account types.Account, defaultPath string) ([]byte, error) {
	// Method implementation...
}
`

Please follow this pattern. cc: @ashish10677

PR #1200, opened 2024-04-05
refactor: used the same keystore instance to fetch pKey everytime 

Yashk767:
# Description

Creating a new keystore instance for every transaction costed us an opening of separate goroutine which would start a watcher/updater to check the changes in the keystore directory files. Either creating the instance at the start and using the same instance every time  for a transaction would function the same and wouldnt cost up an extra goroutine.

- Created keystore instance at the start and reused the same for every transaction.
- Keystore defined as a field to `AccountManager` struct which is used as a method caller to `GetPrivateKey()`

Fixes https://github.com/razor-network/oracle-node/issues/1197

# How Has This Been Tested?

Ran a staker making transactions in voting process with a goroutine counter placed before and after the getPrivateKey call.


cmd/resetUnstakeLock.go
```diff
@@ -50,18 +51,22 @@ func (*UtilsStruct) ExecuteExtendLock(flagSet *pflag.FlagSet) {
...
 	stakerId, err := razorUtils.AssignStakerId(flagSet, client, address)
 	utils.CheckError("Error in getting stakerId: ", err)
 
 	extendLockInput := types.ExtendLockInput{
-		Address:  address,
-		Password: password,
 		StakerId: stakerId,
+		Account:  account,
 	}
-	log.Debugf("ExecuteExtendLock: Calling ResetUnstakeLock with arguments extendLockInput = %+v", extendLockInput)
```
ashish10677, line 64: Why is this log removed.

cmd/vote_test.go
```diff
@@ -275,8 +277,9 @@ func TestCalculateSecret(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			InitializeInterfaces()
-			gotSignature, gotSecret, err := cmdUtils.CalculateSecret(types.Account{Address: tt.args.address,
-				Password: tt.args.password}, tt.args.epoch, testKeystorePath, tt.args.chainId)
+			accountManager := accounts.NewAccountManager(testKeystorePath)
+			account := accounts.InitAccountStruct(tt.args.address, tt.args.password, accountManager)
+			gotSignature, gotSecret, err := cmdUtils.CalculateSecret(account, tt.args.epoch, testKeystorePath, tt.args.chainId)
```
ashish10677, line 282: I think the error check should be first, only then gotSignature and gotSecret must be converted to string.

ashish10677 (changes requested):

cmd/claimBounty.go
```diff
@@ -62,12 +68,10 @@ func (*UtilsStruct) ExecuteClaimBounty(flagSet *pflag.FlagSet) {
 		log.Debug("ExecuteClaimBounty: BountyId: ", bountyId)
 
 		redeemBountyInput := types.RedeemBountyInput{
-			Address:  address,
-			Password: password,
 			BountyId: bountyId,
+			Account:  account,
 		}
 
-		log.Debugf("ExecuteClaimBounty: Calling ClaimBounty() with arguments redeem bounty input: %+v", redeemBountyInput)
```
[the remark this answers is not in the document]
Yashk767: `reedemBountyLockInput` struct has a password field in it so removed it from the logs while other parameters we do log inside `ClaimBounty()`. Same is being done in other scenarios.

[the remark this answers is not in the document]
ashish10677: Okay makes sense.

cmd/vote_test.go
Yashk767: Done in https://github.com/razor-network/oracle-node/pull/1200/commits/9218d2f3fe64728fd2b022f045798ceab67258e4

ashish10677 (approved):

rajkharvar (approved):

PR #1201, opened 2024-04-11
refactor: used a common http client for all the requests

### LINKED ISSUE #1197: Number of goroutines were increasing gradually

Yashk767:
# Description

There was a gradual increase in number of go routines for every http request that node was making, this was due to new HTTP client being used on every `httpClient.Do()` call. Rather there should be a common HTTP client which handles creating and making requests across all hosts, this will enable the reusing of existing connections from the connection pool.
   
- Used a common http client to avoid opening a new connection on every `client.Do()` call for every API.
- Removed global variable `HttpTimeout` and directly picked it up from config.

Fixes #1197 ,  https://linear.app/interstellar-research/issue/RAZ-771

# How Has This Been Tested?

Ran a node on local and staging environment with 5 min epoch time.


core/constants.go
```diff
@@ -81,3 +82,8 @@ var HexArrayExtractIndexRegex = `^hexArray\[(\d+)\]$`
 var BatchSize = 1000
 var NumRoutines = 10
 var MaxIterations = 10000000
+
+// Following are the constants used in custom http.Transport configuration for the common HTTP client that we use for all the requests
+
+var HTTPClientMaxIdleConns = 15
```
ashish10677, line 88: Use const instead of var for these

ashish10677 (changes requested):

core/constants.go
Yashk767: okay I will make this change.
While all the other constants defined in `constants.go` are also defined with `var` keyword. I think we can update to `const` at all these occurrences but in separate PR, not in this. Is this okay @ashish10677 ?

ashish10677: For others, you can open a new PR. For these new variables you've added, change it in this PR itself.

Yashk767: Done in https://github.com/razor-network/oracle-node/pull/1201/commits/897d867f12ee4075cb34e9934e42afacd7518d9c

ashish10677 (approved):

rajkharvar (approved):

ashish10677 (approved):

Issue #1199, opened 2024-03-22
`GetLatestBlockWithRetry` and `GetStakerId` contract call is called repeatedly


Yashk767:
# Description

The RPC usage metrics revealed that calls to `eth_getBlockByNumber` were excessively high. This function is being invoked repeatedly within a for loop during our voting process. To prevent reaching the free tier limit of RPC usage, there's a necessity to optimize this implementation.

Staker Id remains the same for a staker throughout but we still call getStakerId for an address every time inside  `HandleBlock`.

# Solution
We do use `GetLatestBlockWithRetry` in functions  `GetBufferedState` and `GetEpoch` to get the latest block number, rather we can use the existing latest block number provided in 
```
HandleBlock(client *ethclient.Client, account types.Account, blockNumber *big.Int, config types.Configurations, rogueData types.Rogue, backupNodeActionsToIgnore []string)
```

For `stakerId`
 we can just call `GetStakerId(address)` only once outside the voting for loop and use the same when required.

PR #1204, opened 2024-04-19
refactor: Optimised number of `GetBlockNumber` and `GetStakerId` eth calls

Yashk767:
# Description

Reduced number of RPC calls like `GetBlockNumber` and `GetStakerId` which were redundantly called in few functions.

- Removed `GetLatestBlockWithRetry()` calls in functions like `GetBufferedState` and `GetEpoch`and used the existing header provided from the parent `HandleBlock()` function.

- Called `GetStakerId` once and passed the stakerId as a dependency to all the dependent functions as the value of staker Id once received would never change.
 
Fixes # https://github.com/razor-network/oracle-node/issues/1199

### DIFF
--- a/block/block.go
+++ b/block/block.go
@@ -32,9 +32,9 @@ func CalculateLatestBlock(client *ethclient.Client) {
 			latestHeader, err := client.HeaderByNumber(context.Background(), nil)
 			if err != nil {
 				logrus.Error("CalculateBlockNumber: Error in fetching block: ", err)
-				continue
+			} else {
+				SetLatestBlock(latestHeader)
 			}
-			SetLatestBlock(latestHeader)
 		}
 		time.Sleep(time.Second * time.Duration(core.BlockNumberInterval))
 	}
--- a/cmd/cmd-utils.go
+++ b/cmd/cmd-utils.go
@@ -23,7 +23,12 @@ func (*UtilsStruct) GetEpochAndState(client *ethclient.Client) (uint32, int64, e
 	if err != nil {
 		return 0, 0, err
 	}
-	state, err := razorUtils.GetBufferedState(client, bufferPercent)
+	latestHeader, err := clientUtils.GetLatestBlockWithRetry(client)
+	if err != nil {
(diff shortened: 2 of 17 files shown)


utils/api_test.go
```diff
@@ -108,7 +108,7 @@ func TestGetDataFromAPI(t *testing.T) {
 			args: args{
 				urlStruct: types.DataSourceURL{
 					Type:   "POST",
-					URL:    "https://rpc.ankr.com/polygon_mumbai",
+					URL:    "https://sepolia.optimism.io",
```
Yashk767, line 111: Polygon mumbai public provider was down so these tests were failing. Updated it with another testnets op-sepolia network public provider for testing post request to fetch chain Id @ashish10677

block/block.go
```diff
@@ -32,9 +32,9 @@ func CalculateLatestBlock(client *ethclient.Client) {
 			latestHeader, err := client.HeaderByNumber(context.Background(), nil)
 			if err != nil {
 				logrus.Error("CalculateBlockNumber: Error in fetching block: ", err)
-				continue
```
Yashk767, line 35: Removed the `continue` keyword to ensure that the function does not rapidly and repeatedly call `HeaderByNumber` without any pause in case of consecutive failures. 

Removing `conitnue` keyword in case of failure will allow to reach existing sleep function which will be a delay between iterations preventing an excessive number of failed RPC calls.

ashish10677:
Why has the coverage decreased @Yashk767 ?

Yashk767:
> Why has the coverage decreased @Yashk767 ?

Some part of code which was uncovered in coverage is changed in this PR thus it got added it `new missed lines` in coverage resulting in decrease of coverage.
Added new tests for uncovered code, coverage shall increase.

utils/api_test.go
ashish10677: Ok

ashish10677 (approved):

rajkharvar (approved):

PR #1207, opened 2024-04-24
refactor: fetching jobs and collections from asset caches

Yashk767:
# Description

- Stored jobs and collections in a cache and updating the cache on every collection/job update or creation.
- Faster fetching [O(1)] of jobs from cache for every collection/job resulting in quicker execution of commit process.

Fixes https://linear.app/interstellar-research/issue/RAZ-713

# How Has This Been Tested?

Tested with running a staker and creating create and update jobs/collections transactions and checked whether event listeners are listening correctly and updating the cache with the expected update value.


ashish10677 (changes requested):

cmd/eventListeners.go
```diff
@@ -0,0 +1,175 @@
...
+	logs, err := client.FilterLogs(context.Background(), query)
+	if err != nil {
+		log.Errorf("Error in filter logs: %v", err)
+		return []Types.Log{}, nil
+	}
+
+	return logs, nil
+}
+
+func scheduleResetCache(client *ethclient.Client, jobsCache *cache.JobsCache, collectionsCache *cache.CollectionsCache) {
```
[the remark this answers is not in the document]
Yashk767: Hmm yes we don't store any time sensitive data though in the asset caches, as the data stored is completely event-driven and updated regularly in the desired interval, I think we can avoid resetting. correct ?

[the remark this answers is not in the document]
ashish10677: Yes. It's not required.

cmd/eventListeners.go
```diff
@@ -0,0 +1,175 @@
...
+		Addresses: []common.Address{
+			common.HexToAddress(core.CollectionManagerAddress),
+		},
+	}
+
+	// Retrieve the logs
+	logs, err := client.FilterLogs(context.Background(), query)
+	if err != nil {
+		log.Errorf("Error in filter logs: %v", err)
+		return []Types.Log{}, nil
```
ashish10677, line 149: You're not returning error here @Yashk767

cmd/commit.go
```diff
@@ -139,7 +139,7 @@ func (*UtilsStruct) HandleCommitState(client *ethclient.Client, epoch uint32, se
 	log.Debug("HandleCommitState: SeqAllottedCollections: ", seqAllottedCollections)
 	log.Debug("HandleCommitState: Leaves: ", leavesOfTree)
 
-	localCache.StopCleanup()
+	commitParams.LocalCache.StopCleanup()
```
ashish10677, line 142: Why do you need to call StopCleanup() here? It will wait for the cache goroutine to finish right?
But you've just initalized it in this function? Do you need to to start the cache and end it here itself?
I didn't understand this implementation. Please explain.

ashish10677 (changes requested):

cmd/commit.go
Yashk767: No no, the `localCache` described here is the one which stores the result data from APIs. This has nothing to do with caches which we are using for assets.
Here the only change is `localCache` was passed as parameter to function earlier, but now `commitParams`struct instead is passed as a parameter which sub includes the field  `localCache`.

Thus the method caller changes from `localCache.StopCleanup()` to `commitParams.LocalCache.StopCleanup()`.

cmd/eventListeners.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	for _, eventName := range eventNames {
+		eventID := contractABI.Events[eventName].ID.Hex()
+		for _, vLog := range logs {
+			if len(vLog.Topics) > 0 && vLog.Topics[0].Hex() == eventID {
+				switch eventName {
+				case "JobUpdated", "JobCreated":
+					jobId := utils.ConvertHashToUint16(vLog.Topics[1])
+					updatedJob, err := utils.UtilsInterface.GetActiveJob(client, jobId)
```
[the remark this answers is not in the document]
Yashk767: `jobCreated` and `collectionCreated` events only provide `JobId` and `collectionId` when they are emitted.

cmd/eventListeners.go
```diff
@@ -0,0 +1,135 @@
...
+		return nil, err
+	}
+
+	for _, eventName := range eventNames {
+		eventID := contractABI.Events[eventName].ID.Hex()
+		for _, vLog := range logs {
+			if len(vLog.Topics) > 0 && vLog.Topics[0].Hex() == eventID {
+				switch eventName {
+				case "JobUpdated", "JobCreated":
+					jobId := utils.ConvertHashToUint16(vLog.Topics[1])
```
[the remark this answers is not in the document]
Yashk767: Done in https://github.com/razor-network/oracle-node/pull/1207/commits/7d767b2b6b4dbc7e3e9c52a22b465bfdd6f84fee

cmd/eventListeners.go
```diff
@@ -0,0 +1,135 @@
...
+
+// CheckForJobAndCollectionEvents checks for specific job and collections event that were emitted.
+func CheckForJobAndCollectionEvents(client *ethclient.Client, commitParams *types.CommitParams) error {
+	collectionManagerContractABI, err := abi.JSON(strings.NewReader(bindings.CollectionManagerMetaData.ABI))
+	if err != nil {
+		log.Errorf("Error in parsing collection manager contract ABI: %v", err)
+		return err
+	}
+
+	eventNames := []string{"JobUpdated", "CollectionUpdated", "CollectionActivityStatus", "JobCreated", "CollectionCreated"}
```
Yashk767, line 53: use these from constants

core/types/vote.go
```diff
@@ -62,3 +65,11 @@ type ProposeFileData struct {
 	RevealedCollectionIds []uint16
 	RevealedDataMaps      *RevealedDataMaps
 }
+
+type CommitParams struct {
+	JobsCache                 *cache.JobsCache
+	CollectionsCache          *cache.CollectionsCache
+	LocalCache                *cache.LocalCache
+	HttpClient                HttpClientInterface
```
Yashk767, line 73: Cant we directly use inbuilt *client.HttpClient instance instead of this custom HttpClient struct

Yashk767: Done in https://github.com/razor-network/oracle-node/pull/1207/commits/f4514a70fffad796b4103a60679d17931bbdd6f8
@ashish10677 @rajkharvar

ashish10677 (approved):

rajkharvar (approved):

Issue #1208, opened 2024-05-03
Update default config values to support 20 min epoch for staging


Yashk767:
### Description
Currently the `develop` branch has config values set to support 5 min epoch and the same values might not work for 20 min epoch, rather they should be set to support a larger 20 min epoch by default. If the default values are set to support 20 min epoch, they would also support 5 min shorter epoch if the config values are set correctly by the user.

### Implementation 
 Update the default, maximum and minimum config values in `core/constants.go` and cmd/config-utils.go` should be updated to support 20 min epoch.

PR #1209, opened 2024-05-08
fix: updated default config values to support both 20 min and shorter 5 min epoch 

Yashk767:
# Description

Updated the default, maximum and minimum config values in core/constants.go and cmd/config-utils.go` in order to support 20 min epoch which in place by default also supports shorter epoch of 5 min.

Fixes https://github.com/razor-network/oracle-node/issues/1208

### DIFF
--- a/config.sh
+++ b/config.sh
@@ -19,15 +19,15 @@ then
   GAS_MULTIPLIER=1.0
 fi
 
-read -rp "Buffer Percent: (0) " BUFFER
+read -rp "Buffer Percent: (20) " BUFFER
 if [ -z "$BUFFER" ];
 then
-  BUFFER=0
+  BUFFER=20
 fi
 
-read -rp "Wait Time: (1) " WAIT_TIME
+read -rp "Wait Time: (5) " WAIT_TIME
 if [ -z "$WAIT_TIME" ]; then
-   WAIT_TIME=1
+   WAIT_TIME=5
 fi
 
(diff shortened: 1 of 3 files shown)


config.sh
```diff
@@ -19,15 +19,15 @@ then
   GAS_MULTIPLIER=1.0
 fi
 
-read -rp "Buffer Percent: (0) " BUFFER
+read -rp "Buffer Percent: (20) " BUFFER
```
ashish10677, line 22: 20% buffer for 5 minutes is 1 minute in the beginning and 1 minute in the end. So that gives only 3 minutes to process things, and then contract buffer is also there on top of it. Is it okay?

ashish10677 (changes requested):

config.sh
Yashk767: No, this PR just brings back the 20 min support, so only the default, max values are changed for config parameters to support both 20 min and 5 min.
The dev release should support 20 min epoch by default so the default config values are changed for the same.

ashish10677: Okay, so if you want to run 5 min epoch, you'll change the values in your razor.yaml and run right?

Yashk767: yes, the staker would set the config values accordingly for 5 min epoch and the desired input would also lie in the given range limit.

ashish10677 (approved):

rajkharvar (approved):

PR #1214, opened 2024-05-30
chore: updated deployment and default config parameters to support Europa 20 min epoch contracts

Yashk767:
# Description

Current `develop` branch has deployment parameters set for `5 min epoch contracts` on `europa` network, need to update the default contracts to `20 min epoch contracts` on the same network.

This PR Changes, 
- The contract addresses are updated to deployed oracle contracts which has epoch set to 20 min.
- The state length is updated to `1200` sec.  
- This PR also updates the default, max, min config values to support both default 20 min epoch and shorter epochs as mentioned in issue #1208


Yashk767:
Closing this PR as these contains the 20 min staging deployment info, can be reopened when feature PR is ready and merged to develop.

PR #1218, opened 2024-06-19
feat: Implemented batching of multiple getStakeSnapshot RPC calls

### LINKED ISSUE #1217: Batch `getStakeSnapshot` contract calls in calculating biggest staker Id
## Description

Currently in calculating biggest staker node iterates over all the stakers and makes a ` getStakeSnapshot` RPC contract call each time to get the stake of the staker in the current epoch and then returns the staker with biggest stake. 

Making this contract call in each iteration adds on time in overall execution process of calculating biggest staker therefore these `getStakeSnapshot` calls can be batched together and executed as a single RPC batch call which will provide stake values for all the stakers at once which would save a lot of time.

Implement a RPC batch call implementation to get stake for all the stakers at once.

Yashk767:
# Description

- Implemented a generic batch call function where contract address, contract abi and function string input is expected as paramter.
- Used the same generic batch call function to get stakeSnapshot values for all the stakers in calculating biggest staker.
- Added tests for each error case for performing batch calls.  

Fixes #1217

# How Has This Been Tested?

Ran a staker on staging and compared the biggest staker stake and respective staker Id calculated from the current serial implementation of RPC calls.

- Also added tests to get the biggest staker and associated Id.


cmd/propose.go
```diff
@@ -494,6 +501,102 @@ func (*UtilsStruct) GetSmallestStakeAndId(client *ethclient.Client, epoch uint32
 	return smallestStake, smallestStakerId, nil
 }
 
+func (*UtilsStruct) BatchGetStakeSnapshotCalls(client *ethclient.Client, epoch uint32, numberOfStakers uint32) ([]*big.Int, error) {
+	voteManagerABI, err := utils.ABIInterface.Parse(strings.NewReader(bindings.VoteManagerMetaData.ABI))
```
ashish10677, line 505: You can get the ABI just by doing, `bindings.VoteManagerMetaData.GetAbi()`

cmd/propose.go
```diff
@@ -494,6 +501,102 @@ func (*UtilsStruct) GetSmallestStakeAndId(client *ethclient.Client, epoch uint32
...
+	log.Debugf("Stake Snapshot Array: %+v", stakeArray)
+	return stakeArray, nil
+}
+
+func (*UtilsStruct) CreateGetStakeSnapshotBatchCalls(voteManagerABI abi.ABI, epoch uint32, numberOfStakers uint32) ([]rpc.BatchElem, error) {
+	var calls []rpc.BatchElem
+
+	for i := 1; i <= int(numberOfStakers); i++ {
+		log.Debugf("Adding GetStakeSnapshot for Staker Id %v to the batch calls", i)
+		data, err := voteManagerABI.Pack("getStakeSnapshot", epoch, uint32(i))
```
ashish10677, line 539: Add "getStakeSnapshot" to constants

ashish10677 (changes requested):

cmd/propose.go
Yashk767: If I use `GetAbi()` from bindings I wont be able to mock it and cover a case of error in parsing voteManager Abi therefor I just used our existing `AbiInterface.Parse()` which does the same function `abi.Abi()` internally.

cmd/propose.go
Yashk767: Done in https://github.com/razor-network/oracle-node/pull/1218/commits/b7f79ae4f81ca23e0eb97e82c6acfca751d5e296

Yashk767:
Updated the target branch from `develop` to new rebased `feature/v1.2.0` branch.
So there was a need to rebase the base branch.
cc @ashish10677 @rajkharvar

ashish10677:
I think we should implement batch calls with retry @Yashk767 to avoid issues in future.

Yashk767:
> I think we should implement batch calls with retry @Yashk767 to avoid issues in future.

yes okay makes sense, we can have a generic batch function with retry implemented in the same way we did at other place.  @ashish10677

Yashk767:
> I think we should implement batch calls with retry @Yashk767 to avoid issues in future.

Implemented with covering all test scenarios in https://github.com/razor-network/oracle-node/pull/1218/commits/08edf3d85a6a2d35cd42aa85f519c52703e10814
cc @ashish10677 @rajkharvar

ashish10677 (changes requested):

cmd/propose.go
```diff
@@ -224,24 +225,27 @@ func (*UtilsStruct) GetBiggestStakeAndId(client *ethclient.Client, address strin
...
+		return nil, 0, err
+	}
+
+	log.Debugf("Stake Snapshot Array: %+v", stakeSnapshotArray)
 	log.Debug("Iterating over all the stakers...")
 loop:
-	for i := 1; i <= int(numberOfStakers); i++ {
+	for i := 0; i < len(stakeSnapshotArray); i++ {
 		select {
 		case <-stateTimeout.C:
```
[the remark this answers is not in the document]
Yashk767: I think yes, we can remove it now. we just iterate over the stakeArray list of len(numOfStakers).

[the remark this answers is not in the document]
Yashk767: Done in https://github.com/razor-network/oracle-node/pull/1218/commits/7fd941d6bc64b77c338da5296c5771d4d6592da1 @ashish10677

ashish10677 (approved):

rajkharvar (approved):

PR #1221, opened 2024-07-31
refactor: updated default logFileMaxAge and logFileMaxBackups

### LINKED ISSUE #1219: Update the default value for maximum age of log file
### Description

The backups of log files get deleted after `logFileMaxAge` days which is a config parameter.
Currently the default value of `logFileMaxAge` is set to 365 days which equates to 1 year and these backup logs might not be required for most of the stakers for this long period.
These multiple backup logs file staying for a long period on system may pile up and fill good amount of space on the system which might cause memory issues.

So the default value for logFileMaxAge should be changed to 60 days equating to 2 months which is sufficient for a log backup file to be on the system

Also there will be a need to update `logFileMaxBackups` in a way that the number of back up files should be in limit of 2 months expiry time,
so currently `logFileMaxSize` is 200 MB which gets filled in a span of  arounf 7-8 days, thus we have 4 log files of 200MB every month which makes to around 8 log files of 200MB each in span of 60 days.

So we can assign `logFileMaxBackups =10` which is slightly higher number than 8 and justifiable.

Yashk767:
# Description

The multiple backup logs file staying for a long period on system were piling up and filling good amount of space on the system which might cause memory issues, so there was a need to reduce `logFileMaxAge` and `logFileMaxBackups`.

Fixes #1219

# Changes

Updated the default config parameters to 
```
DefaultLogFileMaxBackups = 10
DefaultLogFileMaxAge     = 60
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ docker exec -it razor-go razor setConfig --provider <rpc_provider> --alternatePr
 Example:
 
 ```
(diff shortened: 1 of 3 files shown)


ashish10677 (approved):
lgtm

rajkharvar (approved):

PR #1228, opened 2024-08-29
refactor: added validation on buffer percent limits

Yashk767:
# Description

Added a validation before voting process starts to check buffer percent should be within limit of default value and maximum value.
- While setting buffer percent, it should not be less than default buffer percent value(5) else throw an error.
- While fetching buffer percent during voting, it should be less than max calculated buffer else throw an error.

Maximum value for buffer percent is calculated using state buffer and state length.

Fixes https://linear.app/interstellar-research/issue/RAZ-971
This will fix https://linear.app/interstellar-research/issue/RAZ-965 as well as we update the default buffer percent value for shorter epoch value.

### DIFF
--- a/cmd/config-utils.go
+++ b/cmd/config-utils.go
@@ -424,7 +424,7 @@ func calculateMaxBufferPercent(stateBuffer, stateLength uint64) int32 {
 	}
 
 	// The formula is derived from the condition:
-	// 2(maxBuffer % stateBuffer) < (stateLength - 2*StateBuffer)
+	// 2(maxBuffer % stateLength) < (stateLength - 2*StateBuffer)
 
 	// Perform the calculation with float64 for precision
 	maxBufferPercent := 50 * (1 - (float64(2*stateBuffer) / float64(stateLength)))
--- a/cmd/config-utils.go
+++ b/cmd/config-utils.go
@@ -3,6 +3,7 @@ package cmd
 
 import (
 	"errors"
+	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/sirupsen/logrus"
 	"razor/client"
 	"razor/core"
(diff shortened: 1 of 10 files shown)


cmd/config-utils.go
```diff
@@ -413,3 +403,30 @@ func setLogLevel(config types.Configurations) {
...
+}
+
+// calculateMaxBuffer calculates the maximum buffer percent value.
+func calculateMaxBufferPercent(stateBuffer, stateLength uint64) int32 {
+	if stateLength == 0 {
+		return 0
+	}
+
+	// The formula is derived from the condition:
+	// 2(maxBuffer % stateBuffer) < (stateLength - 2*StateBuffer)
```
ashish10677, line 427: How did you come up with this formula?

ashish10677 (dismissed):

cmd/config-utils.go
Yashk767: total maximum buffer possible should be less than state length:
```
2(StateBuffer) + 2(maxBuffer % StateLength) < Statelength
```

StateBuffer is buffer from contracts and maxBuffer is maximum buffer percent from nodes end.

my bad in comments it should be 
```
// The formula is derived from the condition:
// 2(maxBuffer % stateLength) < (stateLength - 2*StateBuffer)
```	

Its a typo, wrote `stateBuffer` in `2(maxBuffer % stateBuffer)` instead of `stateLength` but used it correctly in formula. Will update the comment.

cmd/config-utils.go
```diff
@@ -207,28 +208,17 @@ func (*UtilsStruct) GetMultiplier() (float32, error) {
...
-	}
-
-	// If bufferPercent is 0, use the default value.
-	if bufferPercentInt32 == 0 {
-		log.Debugf("BufferPercent is unset or set to 0, using its default %d value", core.DefaultBufferPercent)
-		return core.DefaultBufferPercent, nil
+	// bufferPercent cannot be less than core.DefaultBufferPercent else through an error
+	if bufferPercentInt32 < core.DefaultBufferPercent {
+		log.Infof("BufferPercent should be greater than or equal to %v", core.DefaultBufferPercent)
+		return core.DefaultBufferPercent, errors.New("invalid buffer percent")
```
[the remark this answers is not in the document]
Yashk767: For all the config parameters if there's an error we return respective configs default value and error and yes we do process the error if its not nil, so followed the same here.

ashish10677 (approved):

ashish10677 (approved):

PR #1239, opened 2024-09-13
ci: check workflows getting triggered for update artifact actions to latest versions in GitHub workflow

Yashk767:


Yashk767:
Had created this PR to check workflows getting triggered or not from a different branch, we don't need it now. Closing it @yohanelly95

PR #1251, opened 2024-10-01
fix: enabled setting 0 commission for updateCommission command

Yashk767:
# Description

Commission can be set to 0 in update commission transaction from contracts.
So this PR removes the existing commission zero check in `updateCommission` command which will enable the node now to set 0 commission via `updateCommission` command.

- Made the `commission` flag required which will make mandatory to pass commission value now.
- Passing commission as 0 would be accepted for update commission transaction now.

Fixes https://linear.app/interstellar-research/issue/RAZ-1032

### DIFF
--- a/cmd/updateCommission.go
+++ b/cmd/updateCommission.go
@@ -89,7 +89,7 @@ func (*UtilsStruct) UpdateCommission(config types.Configurations, client *ethcli
 	}
 	log.Debug("UpdateCommission: Maximum Commission: ", maxCommission)
 
-	if updateCommissionInput.Commission == 0 || updateCommissionInput.Commission > maxCommission {
+	if updateCommissionInput.Commission > maxCommission {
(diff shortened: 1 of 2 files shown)


Yashk767:
As per the discussion,
The setDelegationAcceptance function checks for a non-zero commission: https://github.com/razor-network/oracle-contracts/blob/main/contracts/Core/StakeManager.sol#L459.

Therefore, the staker must set a non-zero commission before accepting delegation. 
After accepting delegation, they can still update the commission to 0. However, this leads to a poor user experience, as stakers would need to change the commission back to a non-zero value to start accepting delegation again if they decide to stop accepting delegation in the meantime. 
It's better to not allow setting the commission to 0 in order to simplify the process and improve the user experience.

So currently the oracle node doesnt support setting commission to 0 so we will keep the exisiting changes and closing this PR.
cc @yohanelly95

PR #1253, opened 2024-10-04
refactor: replaced SKL with sFuel as a gas token in logs

Yashk767:
# Description

Currently SKL is being logged as  gas token, SKL is never used as an gas ans sfuel and the SKL token are different.
Replaced SKL with sFuel in logs.

### DIFF
--- a/cmd/vote.go
+++ b/cmd/vote.go
@@ -231,9 +231,9 @@ func (*UtilsStruct) HandleBlock(client *ethclient.Client, account types.Account,
 		return
 	}
 
-	// Warning the staker if sFuel balance is less than 0.001 sFuel
+	// Warning the staker if sFUEL balance is less than 0.001 sFUEL
 	if ethBalance.Cmp(big.NewInt(1e15)) == -1 {
-		log.Warn("sFuel balance is lower than 0.001 sFuel, kindly add more sFuel to be safe for executing transactions successfully")
+		log.Warn("sFUEL balance is lower than 0.001 sFUEL, kindly add more sFUEL to be safe for executing transactions successfully")
 	}
 
 	actualStake, err := utils.ConvertWeiToEth(stakedAmount)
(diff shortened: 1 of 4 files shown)


yohanelly95:
@Yashk767 @ashish10677 the min amount will most probably be 0.1 sfuel since the faucet disperses only 0.1 sfuel. Change it to 0.001 sfuel or so imo

Yashk767:
> @Yashk767 @ashish10677 the min amount will most probably be 0.1 sfuel since the faucet disperses only 0.1 sfuel. Change it to 0.001 sfuel or so imo

True, makes sense

ashish10677 (approved):
lgtm

rajkharvar:
@Yashk767 There are other instances where 'sFUEL' is being used. Could you perform a global search and update them as well?

ashish10677:
> @Yashk767 There are other instances where 'sFUEL' is being used. Could you perform a global search and update them as well?

Using sFuel is fine. Using SKL isn't.

rajkharvar (dismissed):

yohanelly95 (approved):

cmd/vote.go
```diff
@@ -215,9 +215,9 @@ func (*UtilsStruct) HandleBlock(client *ethclient.Client, account types.Account,
 		return
 	}
 
-	// Warning the staker if ETH balance is less than 0.1 ETH
-	if ethBalance.Cmp(big.NewInt(1e17)) == -1 {
-		log.Warn("sFuel balance is lower than 0.1 SKL, kindly add more SKL to be safe for executing transactions successfully")
+	// Warning the staker if sFuel balance is less than 0.001 sFuel
+	if ethBalance.Cmp(big.NewInt(1e15)) == -1 {
```
yohanelly95, line 219: @Yashk767 should be `sFUEL` this is from Skales portal

yohanelly95 (changes requested):

cmd/vote.go
Yashk767: I think its fine as its a variable name `ethBalance` as on oracle node all the variables and functions are related to gas are named with eth  like `var ethBalance`, `func CheckEthBalanceIsZero()`. This should work right?

While in logs and comments we have changed it to sFuel wherever required.

Yashk767: @yohanelly95 does this make sense? elsewhere SKL is changed to sFuel wherever needed in comments and logs.

Yashk767: ohh did you mean to use `sFUEL` instead of `sFuel` ?

yohanelly95: yes use sFUEL instead of sFuel @Yashk767

Yashk767: Done in https://github.com/razor-network/oracle-node/pull/1253/commits/93487e9adf49659f95cf6d37a26813618863a8a1

yohanelly95 (approved):

ashish10677 (approved):

rajkharvar (approved):

PR #1254, opened 2024-10-08
feat: added commitment verification layer for data during reveal after removing waitForBlockCompletion check for voting transactions

Yashk767:
# Description

This PR removes the WaitForBlockCompletion for voting transactions Which required some validations needed to be added to make vote work correctly.

- So there was a need to check whether commitment calculated form the commit data fetched from memory during reveal state should be equal to commitment committed suring commit state.
If this check for commitment is verified thn only the node can use the same data during reveal.

- Even during confirm state, node used to save the lastConfirmed check based on confirm transaction status , so that check is now replaced with checking whether is there a block confirmed in the epoch. 
 

Fixes https://linear.app/interstellar-research/issue/RAZ-1024

# How Has This Been Tested?

- Ran multiple nodes with different scenarios mentioned in the issue.
- Ran happy case scenarion on local and staging both for more than 10 epochs.
- Covered the backup node scenario for the same node running on different machines.


ashish10677 (changes requested):
Please make the requested changes.

ashish10677 (approved):

rajkharvar (approved):

PR #1261, opened 2024-11-21
refactor: removed fatal logs from GetTxnOpts()

Yashk767:
# Description

This PR adds a additional return type `error` to `GetTxnOpts()` which would return error in any of the error case instead of existing fatal.  

Fixes https://linear.app/interstellar-research/issue/RAZ-1103

# How Has This Been Tested?


ashish10677:
@Yashk767 there re conflicts

ashish10677 (approved):

rajkharvar (approved):

